PR #136, opened 2015-03-20
Add RCP_Member->can_access() method and wrapper

tnorthcutt:
Adds an RCP_Member class method for determining if the current
user can access the current content. Also adds a wrapper function
`rcp_user_can_access()` for the method.


tnorthcutt:
This also produces an opportunity to do some refactoring on [`rcp_filter_restricted_content()`](https://github.com/pippinsplugins/Restrict-Content-Pro/blob/release/2.1/includes/content-filters.php#L9) and keep things DRYer; let me know if you'd like me to take a pass at that.

pippinsplugins:
Inside of `RCP_Member`, anytime you use `$user_id`, it should be `$this->ID` to ensure it's using the ID of the current member's instance.

includes/class-rcp-member.php
```diff
@@ -457,4 +457,64 @@ public function has_trialed() {
...
+			if ( rcp_user_has_access( $user_id, $access_level ) ) {
+				$ret = true;
+			} else {
+				$ret = false;
+			}
+		} else {
+			$ret = true;
+		}
+
+		return apply_filters( 'rcp_member_can_access', $ret, $this->ID );
```
pippinsplugins: Let's add `$this` as the third parameter here.

includes/member-functions.php
```diff
@@ -319,6 +319,20 @@ function rcp_user_has_access( $user_id = 0, $access_level_needed ) {
 	return false;
 }
 
+/*
+ * Wrapper function for RCP_Member->can_access()
+ * Returns true if user can access the current content
+*/
+function rcp_user_can_access( $user_id = 0 ) {
```
pippinsplugins: I'd like to see an option `$post_id = 0` parameter added here. If it's empty, then we grab the current post.

includes/class-rcp-member.php
tnorthcutt: @pippinsplugins as the third parameter of the `apply_filters` call? If so, any reason not to just pass `$ret` and `$this` and leave out `$this->ID`?

pippinsplugins: It's too be consistent with the other filters in the class and to make it
easier for developers to access the user ID while also allowing them to
access all other class properties.

tnorthcutt: Works for me! Thanks.

tnorthcutt:
@pippinsplugins I think this is all tidied up and ready to go :+1:

tnorthcutt:
@pippinsplugins need anything else from me on this?

pippinsplugins:
Nope! I just need to test it thorougly.

pippinsplugins:
@tnorthcutt This is great. I think I've found some edge cases where it returns improperly, so I'm going to make a few tweaks to it, but excellent work.

tnorthcutt:
@pippinsplugins awesome. Re: edge cases, I'll keep an eye out but if you think to, please @ me – I'd love to learn how to improve it.

Issue #148, opened 2015-04-24
Error creating Stripe Plans


cklosowski:
If you enable the stripe gateway AFTER you create your subscription levels, the subscription comes up as invalid when attempting to process the payment:
```
An error occurred

Status: 400

Message: No such plan: gold
```

It appears that the request doesn't have the proper currency value to configure the plan:
Request:
```
{
    amount: "2000"
    interval: "year"
    interval_count: "1"
    name: "Gold"
    currency: ""
    id: "gold"
}
```
_Notice the missing value for `currency`_

Response:
```
{
    error: {
        type: "invalid_request_error",
        message: "Invalid currency: . Stripe currently supports these currencies: usd, aed, afn, all, amd, ang, aoa, ars, aud, awg, azn, bam, bbd, bdt, bgn, bif, bmd, bnd, bob, brl, bsd, bwp, bzd, cad, cdf, chf, clp, cny, cop, crc, cve, czk, djf, dkk, dop, dzd, eek, egp, etb, eur, fjd, fkp, gbp, gel, gip, gmd, gnf, gtq, gyd, hkd, hnl, hrk, htg, huf, idr, ils, inr, isk, jmd, jpy, kes, kgs, khr, kmf, krw, kyd, kzt, lak, lbp, lkr, lrd, lsl, ltl, lvl, mad, mdl, mga, mkd, mnt, mop, mro, mur, mvr, mwk, mxn, myr, mzn, nad, ngn, nio, nok, npr, nzd, pab, pen, pgk, php, pkr, pln, pyg, qar, ron, rsd, rub, rwf, sar, sbd, scr, sek, sgd, shp, sll, sos, srd, std, svc, szl, thb, tjs, top, try, ttd, twd, tzs, uah, ugx, uyu, uzs, vnd, vuv, wst, xaf, xcd, xof, xpf, yer, zar, zmw, vef.",
        param: "currency"
    }
}
```

PR #149, opened 2015-04-24
Fixing plans not being created when they do not exist, upon purchase

cklosowski:
Addresses #148


cklosowski:
@pippinsplugins This should take care of that stripe bug I was seeing

Issue #151, opened 2015-04-28
Payment through AJAX instead of reloading page with form post submit?


ahansson89:
I like AJAX and would love to see something like this.

pippinsplugins:
This is already done in `release/2.1` via #120

PR #152, opened 2015-04-28
Submit through ajax instead of form submit

ahansson89:
Was thinking something along the lines of this #151


ahansson89:
First of all it makes it easier to debug when you can get the ajax output of the process function. On the other hand it is nice because customers are not taken away from the payment page and therefore they do not need to re-enter their information if it is typed incorrectly.

pippinsplugins:
@ahansson89 See my comment on #151 :)

ahansson89:
Oh... Awesome!

pippinsplugins:
:)

PR #154, opened 2015-05-03
Simplify and secure data saving in Restrict meta box.

fklein-lu:
- Use `wp_nonce_field()` to output the nonce with a unique name and action.
- Use `sanitize_text_field()` for sanitizing the data.
- Rearrange function logic for more clarity.
- Only attempt to delete meta data generated by checkboxes, as the other options all save default values.


pippinsplugins:
Looks good!

I'll give it a test and merge it in :)

pippinsplugins:
@fklein-lu This needs a check for array values. Right now all values are passed through `sanitize_text_field()`, which results in `Notice: Array to string conversion in /Users/pippin/Sites/edd/wp-includes/formatting.php on line 766` notices since the subscription levels are passed as an array

fklein-lu:
Thanks for testing the PR. I did not run into this notice while testing, my apologies.

But considering this, I think it would be a good idea to have some unit tests for this. I'll try and submit some with an updated PR.

fklein-lu:
Turns out that the project I'm developing does not yet use subscription levels, hence the missed notice.

I created a basic unit test to reproduce the issue, and added an additional check. There are a couple of other tests, trying to verify the behavior that the plugin currently has to avoid regressions.

I find that the way that the tests are set up is inelegant. I've never unit tested a meta box before, so I'm unsure about what would be a more elegant solution. Maybe you have an idea for a better approach.

pippinsplugins:
Did you mean to close this?

PR #169, opened 2015-05-24
Issue/168

### LINKED ISSUE #168: Add labels to subscription levels
I think the subscription levels should have labels so you can easily click on the text and change the level

screen shot 2015-05-24 at 9 04 30 pm

sumobi:
Fix #168.

I haven't used RCP a lot so I may be overlooking something - correct me if I'm wrong.

1. The .rcp_form label CSS class isn't needed. We don't need to float a label or set a fixed width on it. Display block might come in handy still.

2. The list items inside the `register.php` template should have their IDs removed and added as a class instead. This way we can add the IDs to the new label and you'll be able to switch the subscription level by clicking on the text. The class still allows the list item to be styled if need be (unless there's something I've missed).


pippinsplugins:
The label CSS is needed because that's what forces the labels and inputs to be next to each other.

sumobi:
@pippinsplugins The only forms I can see that use this CSS class is the profile editor form:

screen shot 2015-05-25 at 3 47 40 pm

And login form:

screen shot 2015-05-25 at 3 47 13 pm

The subscription details form doesn't seem to be affected by the label width:

screen shot 2015-05-25 at 3 47 34 pm

And of course the registration form looks beautiful as is:

screen shot 2015-05-25 at 3 47 47 pm

If we remove the aforementioned CSS rule, the forms will look nicer. It's also consistent with the register form where the checkbox and radio button are shown first, then the label. Provided though I haven't missed a form, i went through all the shortcodes to test.

Profile editor:
screen shot 2015-05-25 at 3 53 05 pm

Login:
screen shot 2015-05-25 at 3 53 11 pm

And finally, I think we should modify the subscription details form so the checkbox is on the left, then the label:

screen shot 2015-05-25 at 3 57 28 pm

sumobi:
Ah, I see now. The 2015 theme has CSS which overrides RCP's styling and makes input fields 100% width: 
screen shot 2015-05-25 at 6 16 03 pm

I was a bit thrown off because the screenshots also show the 100% width fields:
https://pippinsplugins.com/products/restrict-content-pro/#jp-carousel-60620

So now that I've disabled that CSS file I see the purpose of the fixed label width:

screen shot 2015-05-25 at 6 16 58 pm

IMO I still think it's better to not have a fixed width for labels though I can understand that many users may be used to how it's currently styled so it's definitely a hard one to change. Having the label with input field underneath gives more flexibility and compatibility for mobile devices though it's easy to override if need be.

I'll re-submit without the CSS modifications and leave the other label changes intact :+1:

pippinsplugins:
I'd rather not change it for now, simply because too many sites have based their designs off of the CSS that's there already.

pippinsplugins:
I need to make sure this doesn't interfere with any of the gateway JS before merging but otherwise looks good.

sumobi:
Just adding another comment to show my support for removing fixed widths, sometime in the future. This doesn't look good:

2015:

screen shot 2015-10-04 at 10 08 17 pm

2016:

screen shot 2015-10-04 at 10 08 53 pm

@pippinsplugins new major version with highlighted note about CSS changes? Detect CSS template (similar to EDD) and inject inline CSS?

PR #177, opened 2015-05-27
Add label to subscription levels

ghost1227:
Issue #168


pippinsplugins:
Andrew already did this in #169

PR #187, opened 2015-06-08
Removed Duplicate Code

jawittdesigns:
Check 'status' column in restrict_content_pro table in includes/upgrades.php on line 63 duplicated on line 77. Removed duplicate code from line 77.


pippinsplugins:
Thanks!

PR #188, opened 2015-06-18
fix fieldset name

bearded-avenger:
fixed fieldset class name. ran a search and didn't see the misspelling used anywhere else except in these two files


pippinsplugins:
Thanks!

PR #200, opened 2015-07-02
Feature/recaptcha api v2

jb510:
Incomplete.  New No Captcha ReCaptha using APIv2 is working, but the verification functions pulled over from v1 aren't debugged/working yet.


pippinsplugins:
Could you submit it against the issue/199 branch?

PR #213, opened 2015-08-08
API response Issue

jawittdesigns:
Here is the code being used for the 2Checkout card authorization, so far


pippinsplugins:
Manually merged into `issue/203`

PR #214, opened 2015-08-11
Use is_ssl() vs $_SERVER["HTTPS"] as it works behind most reverse pro…

nciske:
…xies

Fixes an issue where RCP was adding :443 but not using https:// which caused a 400 Bad Request "The plain HTTP request was sent to HTTPS port" error.

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -243,7 +243,7 @@ function rcp_get_current_url() {
 	else :
 
 		$current_url = 'http';
-		if ( isset( $_SERVER["HTTPS"] ) && $_SERVER["HTTPS"] == "on" ) $current_url .= "s";
+		if ( is_ssl() ) $current_url .= "s";
 
 		$current_url .= "://";
 


pippinsplugins:
@nciske when/where was the request failing?

nciske:
During the cancellation process.

The user ended up on; http://domain.com:443/my-account/

Host is WebSynthesis.

Nick Ciske
@nciske

tnorthcutt:
I've seen a similar issue a few times, but didn't think to connect it to RCP - just assumed it was a server issue (which it still could be, in my case).

mindctrl:
I've seen this be a problem once in EDD with `edd_get_current_page_url` ([link](https://github.com/easydigitaldownloads/Easy-Digital-Downloads/blob/b64d4a72856005fce4a6ac786422919760fb49ea/includes/misc-functions.php#L439)). The server wasn't setting `$_SERVER["HTTPS"]`. I like that `is_ssl()` checks the server port as a fallback. If you accept this here can we also do it in EDD?

pippinsplugins:
Going to milestone this for 2.3. I'll test further and merge.

@mindctrl yes let's do the same in EDD.

PR #217, opened 2015-08-17
Add filter points

dcavins:
Hi Pippin-

I've been exploring whether your plugin will work with a BuddyPress installation, and I think it will. I'm requesting the addition of a few filters specifically so that the `registration_page` option can be handled dynamically. In BP, it makes sense for users to visit their profiles to manage their subscriptions, but that URL is different for each user.

The bigger change is moving the creation of the global variables to WP's `init` hook. I didn't see any actions that occurred earlier in the page load cycle, so I hope that's a safe move. The problem was that if the global variables are assigned as the plugin is loaded, other plugins can't reliably filter those values.

Please let me know if you have any questions.

-David


dcavins:
Yeah, so moving the global variable creation to the init hook does cause problems. So please ignore this PR until l I have some time to look into that some more.

PR #228, opened 2015-09-09
Add filters to cron queries

foliovision:
Hello Pippin,

we just added some hooks to make sure we can avoid members using our Payflow gateway from expiring right away. Our code tries to check for a payment again in a week and only then expires the user. 

Thanks,
Martin


pippinsplugins:
@foliovision Seems like a good filter to me. Is there a reason you need a filter on both the query and the results? The query should dictate the results so that seems unnecesary.

foliovision:
@pippinsplugins 

I think having both would be helpful. Sometimes a SQL query might not be enough to pick the members should should be excluded from the process, it might require some PHP processing in complex setups and the "rcp_check_for_expired_users_members_filter" filter would allow that.

Thanks,
Martin

PR #240, opened 2015-09-25
fix thie->this spelling bug on cat page

chuckreynolds:
realized I don't know how to deal with .mo files but figured with this heads up you'd be able to fix that up quickly. Cheers


chuckreynolds:
redoing this fix. closing this.

Issue #233, opened 2015-09-14
Add support for Stripe customer Credit


carter2422:
As it is right now RCP does not seem to support adding credit to customers accounts in Stripe. This would be very useful, though. 

Currently if you want to comp someone's account then you must add a negative value invoice item, such that a payment is still created (and thus updates RCP) but doesn't actually charge the user the full subscription amount. 

The current issue is that if you add Credit to a customer's Stripe account, and that credit is greater than the subscription cost, then Stripe will not insert a payment for that customer until the credit is used up. This causes RCP to not update and thus the customer becomes expired in RCP, even though their subscription is still active.

pippinsplugins:
I'm not sure this is something I can fix.

When a customer has credit greater than their payment amount, Stripe doesn't send the webhook that RCP relies on for recording payments.

I may be able to use a different webhook but I'm not sure.

carter2422:
We should be able to tackle this by checking for `starting_balance` during invoicing. I'll try and tackle.

pippinsplugins:
:+1:

carter2422:
Thanks @pippinsplugins!

PR #267, opened 2015-10-07
Feature/stripe credit

carter2422:
This add support for credit on customers Stripe profiles by first checking for Invoice events, rather than only the charge event. Addresses #233 

It is mostly working except something is amiss in setting of the `transaction_id` and the `amount` in the payment data. Everything is coming through in `$payment_event` just fine, not sure what's going on.


carter2422:
Okay @pippinsplugins, this is fully working now so far as I can tell. All of my tests have succeeded:
- One time charges added to customers account
- Recurring charges via invoice on a subscription
- Recurring charges via invoice where part of `$amount` is covered by credit on customers account
- Recurring charges via invoice where all of the `$amount`is covered by credit and thus no charge is created.

Issue #271, opened 2015-10-08
Refactor rcp_check_for_existing_payment() to use transaction_id


carter2422:
The `rcp_check_for_existing_payment()` function doesn't work reliably and it seems it'd be far more reliable to simply check against transaction IDs. In particular, this would help avoid duplicate payments where the date is different for one reason or another. 

One such example is if an admin has manually entered payments with their transaction IDs after a web hook failure. If that web hook were to be retired it'll currently create duplicate payments as it's comparing the exact date, not the transaction ID.

pippinsplugins:
@carter2422 You can use it like this:
```php
$payments = new RCP_Payments;

if( $payments->payment_exists( $trans_id ) ) {
     // it exists
}
```

carter2422:
Nice. Thanks @pippinsplugins

PR #272, opened 2015-10-08
Introduce payment-functions.php file and function to check for existing payments by ID

carter2422:
This introduces a new `payment-functions.php` file, since all current payment functions are depreciated. In addition this adds a new `rcp_check_for_existiing_payment_by_id()` function that only compares the unique transaction ID to determine if a payment exists in the RCP database already. 

The previous function, depreciated, function `rcp_check_for_existing_payment()` used the `$type`, `$date`, and `$subscription_key` to compare payments. This actually resulted in duplicate (yet incomplete payment) entries in some scenarios. One example is in the work I'm doing for #233, where it would sometimes create up to 6 payment entries, with only the last one being complete.

Addresses #271

### DIFF
--- a/includes/payment-functions.php
+++ b/includes/payment-functions.php
@@ -1,0 +1,18 @@
+<?php 
+/**
+ * Check if a payment already exists
+ *
+ * @access      private
+ * @param       $transaction_id string/date The unique ID associated with the transaction
+ * @return      bool
+*/
+
+function rcp_check_for_existing_payment_by_id( $transaction_id ) {
+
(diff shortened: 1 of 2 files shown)


pippinsplugins:
No need to introduce a new function or file. We can just add a new method to the class. I prefer that over global functions.

pippinsplugins:
I'm going to refactor the existing `payment_exists()` function for this. It's only used internally so it's safe to change.

carter2422:
Great!

PR #288, opened 2015-11-03
Add actions to metabox.php

tannerm:
I am trying to add BuddyPress specific content restrictions and would like to place them after the default fields while still maintaining control over the functionality.


tnorthcutt:
I haven't tested this locally, but from a first glance this gets a :+1: from me.

tannerm:
Thanks @tnorthcutt! I don't expect there's much that can go wrong with a couple of new actions. :)

pippinsplugins:
Whoops, missed this!

PR #315, opened 2016-01-06
Payment failed hooks

jmarx:
Remove all the try/catch error blocks(except for the base) because they are unnecessary and make it needlessly complicated to put in a hook that allows us to react to stripe error. I checked out all the different classes in the stripe library and the behavior for each failure is the same. I also tested and and the behavior for all errors is the same with or without all those pointless error handlers. #315


pippinsplugins:
@jmarx We still need to properly catch each of the different kinds of exceptions so whatever solution we add here needs to properly respect those.

jmarx:
@pippinsplugins fair enough. I will restore the blocks and add a hook for each one. Will have that for you later on.

jmarx:
@pippinsplugins Thoughts on this approach?

pippinsplugins:
I like this. I'll get it merged in for the next update.

Issue #293, opened 2015-11-22
Wrong auto renew date


tannerm:
When a user cancels an auto renew subscription then switches to another auto renew subscription, the new renewal date is incorrect when compared with Stripes renewal date.

To recreate, register for a subscription with auto renew checked. Then cancel the subscription and upgrade to a new one. Once the new subscription receives a pingback from Stripe you'll notice that the renewal date from RCP is different from the renewal date in Stripe.

Here's a screencast:
https://www.dropbox.com/s/nb2m26et9sh1stm/issue.mp4?dl=0

tannerm:
@pippinsplugins This seems like a pretty big deal and want to make sure you see it.

swb1192:
When this renewal goes through, it looks like it is doing a duplicate of the renewal, as per the User Notes in the member data:

November 22, 2015 18:23:19 - Member's expiration changed from 2016-11-22 23:59:59 to 2017-11-22 23:59:59

November 22, 2015 18:23:19 - Member's expiration changed from 2017-11-22 23:59:59 to 2018-11-22 23:59:59

pippinsplugins:
I'll dig into this as soon as I can.

pippinsplugins:
@tannerm How much are the dates off by?

tannerm:
@pippinsplugins normally by a year or two. (I'm assuming that would change to whatever unit is used for the subscription).

swb1192:
The renewals should be 1 year, but instead they are two years (see https://github.com/pippinsplugins/restrict-content-pro/issues/293#issuecomment-159436317).

pippinsplugins:
Pretty sure I've tracked this down.

pippinsplugins:
@tannerm Could you test it against the PR above? I'm pretty sure that resolves it.

pippinsplugins:
#303 should resolve this.

tannerm:
Sorry for the delay here, @pippinsplugins. I just pulled down the master branch and tested again and am still seeing the same results. Happy to do a screenshare if that would help.

tannerm:
Or if you wait long enough, I'll have time to dig into this myself. :P

pippinsplugins:
Ok I'm struggling to find what's causing this because I cannot replicate it at all.

Any ideas from anyone here?

tannerm:
Yeah, I'll look into this. I should have some time over the next couple of days.

pippinsplugins:
[...]

PR #316, opened 2016-01-09
Fix Stripe Subscriptions Update Issue.

tannerm:
Hey @pippinsplugins! This is a pull request for issue #293.

Here is the issue that I found:

A user creates an account with auto-renew (using Stripe subscription) and the new expiration date is set (let's say for 1 year). That user then decides to upgrade, so he cancels and then upgrades to a new account again with auto-renew. 

With the current code, the expiration date calculated is added on to the existing expiration date because the user is activated before the payment is actually processed.

The solution that I have included sets the user to pending if auto-renew is turned on thereby allowing the correct expiration date to be calculated. I also added some modifications to rcp_set_pending_subscription_on_upgrade(). Now when a user is set to pending, the new subscription and expiration date shows in the subscription details shortcode.

Let me know if you have any questions or want me to make any modifications. Also happy to hop on a call if you want to discuss further.


includes/registration-functions.php
```diff
@@ -490,5 +490,12 @@ function rcp_set_pending_subscription_on_upgrade( $status, $user_id ) {
 		delete_user_meta( $user_id, 'rcp_pending_subscription_level' );
 		delete_user_meta( $user_id, 'rcp_pending_subscription_key' );
 	}
+
+	// this expiration date will be updated later. It's for the user's benefit
+	if ( 'pending' == $status ) {
```
pippinsplugins: I'm pretty sure this breaks expiration date calculations in PayPal.

pippinsplugins: I also don't believe this is necessary since the expiration date is calculated when the `renew()` method is called inside the web hook processing.

tannerm: ok... I don't have PayPal setup locally so didn't test with that.

`renew()` does call the expiration date recalculation function, but since stripe previously set the subscription to "active" it gets bypassed.

pippinsplugins:
I've just done some testing and this doesn't seem to work right. I'll test more when not on a plane.

pippinsplugins:
When testing, it wasn't updating the expiration date; it stayed as the same day as registration.

pippinsplugins:
See https://github.com/restrictcontentpro/restrict-content-pro/issues/293#issuecomment-172170414

Issue #318, opened 2016-01-12
Subscription levels can be created without entering any values


pippinsplugins:
The subscription level creation form needs to validation on it. Currently, you can create subscription levels without a name or price. At least the name should be required.

misulicus:
i added this to the admin-scripts.js and the localisation and it should be enough

````
	$('#rcp-member-levels').submit(function() {
		if($('#rcp-name').val() == '') {
			alert(rcp_vars.missing_level_name);
			return false;
		}
		return true;
	});	
	$('#rcp-member-levels').submit(function() {
		if($('#rcp-duration').val() == '') {
			alert(rcp_vars.missing_level_duration);
			return false;
		}
		return true;
	});		
	$('#rcp-member-levels').submit(function() {
		if($('#rcp-price').val() == '') {
			alert(rcp_vars.missing_level_price);
			return false;
		}
		return true;
	});
````

pippinsplugins:
Thanks!

PR #320, opened 2016-01-12
validate empty level fields when creating

misulicus:
Check values for new level creating Part 2 for Fixing #318

### DIFF
--- a/includes/process-data.php
+++ b/includes/process-data.php
@@ -22,7 +22,12 @@ function rcp_process_data() {
 			if( ! current_user_can( 'rcp_manage_levels' ) ) {
 				wp_die( __( 'You do not have permission to perform this action.', 'rcp' ) );
 			}
-
+			
+			if( empty( $_POST['name'] ) || empty( $_POST['duration'] ) || empty( $_POST['price'] ) ) {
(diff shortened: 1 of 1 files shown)


includes/process-data.php
```diff
@@ -22,7 +22,12 @@ function rcp_process_data() {
 			if( ! current_user_can( 'rcp_manage_levels' ) ) {
 				wp_die( __( 'You do not have permission to perform this action.', 'rcp' ) );
 			}
-
+			
+			if( empty( $_POST['name'] ) || empty( $_POST['duration'] ) || empty( $_POST['price'] ) ) {
+				$url = get_bloginfo('wpurl') . '/wp-admin/admin.php?page=rcp-member-levels&rcp_message=level_missing_fields';
```
mindctrl: @misulicus shouldn't this use `admin_url()` instead?

sumobi: :+1: Also, wrap in `esc_url`

misulicus:
````
$url = admin_url( 'admin.php?page=rcp-member-levels&rcp_message=level_missing_fields' );
wp_safe_redirect( esc_url_raw( $url ) ); exit;
`````
Something like that will work ?

PR #321, opened 2016-01-13
validate empty level fields when creating

misulicus:
Check values for new level creating Part 2 for Fixing #318

### DIFF
--- a/includes/process-data.php
+++ b/includes/process-data.php
@@ -23,6 +23,11 @@ function rcp_process_data() {
 				wp_die( __( 'You do not have permission to perform this action.', 'rcp' ) );
 			}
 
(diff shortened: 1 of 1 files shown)


includes/process-data.php
```diff
@@ -23,6 +23,11 @@ function rcp_process_data() {
 				wp_die( __( 'You do not have permission to perform this action.', 'rcp' ) );
 			}
 
+			if( empty( $_POST['name'] ) || empty( $_POST['duration'] ) || empty( $_POST['price'] ) ) {
```
pippinsplugins: Duration is allowed to be left blank for lifetime subscriptios.

Issue #328, opened 2016-01-26
Validate user input on discount code creation screen


mindctrl:
When creating a discount code, if you key in the `%` sign in the amount field, you can see an error 400, invalid integer message.

![image](https://cloud.githubusercontent.com/assets/1254870/12589743/47aa420c-c42d-11e5-92a6-f5bb56485f9a.png)

![image](https://cloud.githubusercontent.com/assets/1254870/12589729/3857e480-c42d-11e5-9370-bab00b32e3a5.png)

Customer reported this problem. I replicated locally on one site. Another site allows it through, but shows a double `%% on the list:

![image](https://cloud.githubusercontent.com/assets/1254870/12589795/7e52abdc-c42d-11e5-8283-92910e06e79e.png)

michaelbeil:
We would clarify the description by saying: "The amount of this discount code. Use a number without % or $, such as 15."

Or we could just strip any non integer characters on output.

mindctrl:
Left a note on the PR here: https://github.com/restrictcontentpro/restrict-content-pro/pull/336#issue-129893846

pippinsplugins:
See comments on PR :)

PR #336, opened 2016-01-29
Validate input during discount code creation and updates. #328.

mindctrl:
#328.

I need some feedback. I'm not sure about this approach, and I'm not aware of what other things might hook into this.

Stripe doesn't allow percentage discounts with a decimal. For example, you can't use 25.5%.

With that in mind, the PR prevents decimals in percentage discounts. If we need to allow them for other reasons, we'll need to do some gateway detection and/or create a map of allowed discount formats for each gateway.

Things this PR addresses:
- if someone fails to enter a discount amount, it throws a notice and does not create the discount.
- checks that percentage discounts are whole numbers between 1 and 100, based on the comment above, and throws a notice if it doesn't fall within the range 
- ensures the amount entered is a valid number

Since this needs to run in at least two places, I created a helper method for formatting the amount based on the unit type.

Thoughts?

### DIFF
--- a/includes/class-rcp-discounts.php
+++ b/includes/class-rcp-discounts.php
@@ -287,6 +287,14 @@ class RCP_Discounts {
 
 		$args = wp_parse_args( $args, $defaults );
 
+		$amount = $this->format_amount( $args['amount'], $args['unit'] );
+
+		if ( is_wp_error( $amount ) ) {
+			wp_die( $amount->get_error_message() );
+		} else {
+			$args['amount'] = $amount;
+		}
+
 		if( $this->get_by( 'code', $args['code'] ) ) {
 			return false; // this code already exists
 		}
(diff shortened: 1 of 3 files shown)


includes/class-rcp-discounts.php
```diff
@@ -287,6 +287,14 @@ public function insert( $args = array() ) {
 
 		$args = wp_parse_args( $args, $defaults );
 
+		$amount = $this->format_amount( $args['amount'], $args['unit'] );
+
+		if ( is_wp_error( $amount ) ) {
+			wp_die( $amount->get_error_message() );
```
pippinsplugins: I don't think we should die here. Since this function could be called programmatically somewhere else, it would be better to just return the WP_Error.

includes/class-rcp-discounts.php
```diff
@@ -342,6 +350,14 @@ public function update( $discount_id = 0, $args = array() ) {
 
 		$args     = array_merge( $discount, $args );
 
+		$amount = $this->format_amount( $args['amount'], $args['unit'] );
+
+		if ( is_wp_error( $amount ) ) {
+			wp_die( $amount->get_error_message() );
```
pippinsplugins: Same thing here. Let's return the WP_Error

includes/class-rcp-discounts.php
pippinsplugins: Then inside the process-data.php file, check the return when creating / updating a discount and call wp_die() there

includes/process-data.php
```diff
@@ -270,6 +270,10 @@ function rcp_process_data() {
 
 			$add = $discounts->insert( $data );
 
+			if ( is_wp_error( $add ) ) {
+				wp_die( $add );
```
pippinsplugins: This shoudl show the error message itself.

PR #345, opened 2016-02-08
Account for letting users resubscribe when they have a previous unpaid subscription in Stripe

### LINKED ISSUE #302: Stripe: account for Unpaid subscriptions when re-subscribing
If a user tries to re-subscribe via Stripe, and has a previously unpaid subscription (typically due to failed payments) then the new subscription never goes through but RCP reports it as successful to the user. 

There needs to be a way to account for unpaid subscriptions in Stripe.

Note: after signup no error messages are presented, so it appears successful, but the user is never actually activated.

carter2422:
This PR fixes #302, whereby a user with a previous, `unpaid` subscription was blocked for signing up again to a new plan. 

The route I've taken works, but it may not be ideal. 

If a user has an `unpaid` subscription then we first cancel that old one and then subscribe them fresh to the new subscription. This works. The problem, though, is that it does not get counted as a *Upgrade* subscription, even though they're technically changing plans. I'm not sure if this is a good thing or a bad thing.

For example, here is the event log for the test user that I used: 
screen shot 2016-02-08 at 5 57 47 pm

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-stripe.php
+++ b/includes/gateways/class-rcp-payment-gateway-stripe.php
@@ -139,6 +139,21 @@ class RCP_Payment_Gateway_Stripe extends RCP_Payment_Gateway {
 
 				}
 
+				// clean up any past due or unpaid subscriptions before upgrading/downgrading
+				foreach( $customer->subscriptions->all()->data as $subscription ) {
+
+					// check if we are renewing an existing subscription. This should not ever be 'active', if it is Stripe
+					// will do nothing. If it is 'past_due' the most recent invoice will be paid and the subscription will become active
+					if ( $subscription->plan->id == $plan_id && in_array( $subscription->status, array( 'active', 'past_due' ) ) ) {
+						continue;
+					}
+
+					// remove any subscriptions that are past_due or inactive
+					if ( in_array( $subscription->status, array( 'past_due', 'unpaid' ) ) ) {
+						$subscription->cancel();
+					}
+				}
+
 				// Save the card and any coupon
 				$customer->save();
 
@@ -471,23 +486,6 @@ class RCP_Payment_Gateway_Stripe extends RCP_Payment_Gateway {
 
 					if( ! $member->get_subscription_id() ) {
 						die( 'no subscription ID for member' );
-					}
-
-					// in case a user attempts to resubscribe but has a previously unpaid subscription
-					if( $event->type == 'customer.subscription.updated' && $payment_event->status == 'unpaid' ) {
-
-						// make sure we're acting on a upgrade or downgrade only
-						if( $event->data->previous_attributes->plan ) {
-
-							// cancel the previous plan
-							$customer = \Stripe\Customer::retrieve( $payment_event->customer );
-							$customer->subscriptions->retrieve( $payment_event->plan->id )->cancel();
-
-							// add the new plan
-							$customer->updateSubscription( array( 'plan' => $payment_event->id ) );
-
-						}
-
 					}
 
 					if( $event->type == 'charge.succeeded' || $event->type == 'invoice.payment_succeeded' ) {
--- a/includes/gateways/class-rcp-payment-gateway-stripe.php
+++ b/includes/gateways/class-rcp-payment-gateway-stripe.php
@@ -473,6 +473,23 @@ class RCP_Payment_Gateway_Stripe extends RCP_Payment_Gateway {
 						die( 'no subscription ID for member' );
 					}
 
+					// in case a user attempts to resubscribe but has a previously unpaid subscription
+					if( $event->type == 'customer.subscription.updated' && $payment_event->status == 'unpaid' ) {
+
+						// make sure we're acting on a upgrade or downgrade only
+						if( $event->data->previous_attributes->plan ) {
+
+							// cancel the previous plan
+							$customer = \Stripe\Customer::retrieve( $payment_event->customer );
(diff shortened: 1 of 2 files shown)


pippinsplugins:
Does changing their plan automatically bill them as well?

carter2422:
Yes. Adding the plan also handles the first invoice and payment.

pippinsplugins:
@tannerm Could you look at this as well in next couple of days?

tannerm:
@pippinsplugins Yeah, I'll look at it right now.

tannerm:
Alright, I'm running some tests on this, I'll let you both know what I find.

pippinsplugins:
@carter2422 Do you have this running on the live site?

carter2422:
I had it on out development server for testing. It's not on live yet. 

Jonathan Williamson
http://cgcookie.com

pippinsplugins:
Tanner has a test setup that should finalize tomorrow I believe. We'll merge after that if all is good.

carter2422:
:+1:

tannerm:
Still looking into this... just gotta wait for everything to, you know, expire. ಠ__ಠ

carter2422:
@tannerm just in case you haven't seen it this works well for testing this: http://stackoverflow.com/questions/24815799/how-can-i-test-unpaid-subscriptions-in-stripe-with-ruby-on-rails

Just requires one day delay for testing.

tannerm:
awesome, thanks @carter2422!

tannerm:
@pippinsplugins @carter2422 I just committed a change for this. @carter2422 I wasn't able to test the code that you wrote as I never had webhook with the right parameters (not sure why). But the code that I've added addresses this issue and also solves another issue you've noted where renewing users are double charged.

includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -139,6 +139,21 @@ public function process_signup() {
 
 				}
 
+				// clean up any past due or unpaid subscriptions before upgrading/downgrading
+				foreach( $customer->subscriptions->all()->data as $subscription ) {
+
+					// check if we are renewing an existing subscription. This should not ever be 'active', if it is Stripe
+					// will do nothing. If it is 'past_due' the most recent invoice will be paid and the subscription will become active
+					if ( $subscription->plan->id == $plan_id && in_array( $subscription->status, array( 'active', 'past_due' ) ) ) {
```
pippinsplugins: Wouldn't this skip a past_due subscription so the cancellation below is never called?

includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -473,6 +488,23 @@ public function process_webhooks() {
 						die( 'no subscription ID for member' );
 					}
 
+					// in case a user attempts to resubscribe but has a previously unpaid subscription
+					if( $event->type == 'customer.subscription.updated' && $payment_event->status == 'unpaid' ) {
```
pippinsplugins: Minor consistency note here: we use yoda conditions, so `unpaid` should come first.

tannerm: Oh, this whole section was supposed to go away. Must have gotten re-introduced with the merge conflict. Just removed.

includes/gateways/class-rcp-payment-gateway-stripe.php
tannerm: @pippinsplugins only if the subscription that is past due is the same subscription the user is trying to renew. Basically, this check allows us to handle a user who's subscription expired and they chose to renew. This maintains the same billing schedule etc.

pippinsplugins: The scenario makes sense but I don't see how a `past_due` hook ever gets past this first time and reaches the cancel() call below. Any subscription that has a status of `past_due` will be stopped by this first conditional.

tannerm: Are you seeing the `$subscription->plan->id == $plan_id`? So any plan that is not the plan the user is upgrading/downgrading fail this conditional.

So if a user has a subscription (call it "Silver") that is expired (past_due or unpaid) and he is upgrading to a new subscription (call it "Gold") this loop will see the Silver plan, check and make sure it is not the Gold plan (the plan the user is upgrading to), and then cancel it since it is expired.

Alternatively, say that same user has the Silver plan, again it expired (say credit card issue) and they want to renew the plan. That conditional will short circuit this functionality since the subscription the user is checking out with is the same they currently have. Thus avoiding a new plan renewal date.

Let me know if I'm missing something. Or just overcomplicating this. :)

carter2422:
I'll try and test this tomorrow against my testing users.

includes/gateways/class-rcp-payment-gateway-stripe.php
pippinsplugins: It's very likely I am the one over complicating it :)

I'll look again in the morning when I have fresh eyes.

pippinsplugins:
@carter2422 @tannerm Were either of you able to confirm this fixed it?

carter2422:
@pippinsplugins I can test later today or early next week.

includes/gateways/class-rcp-payment-gateway-stripe.php
pippinsplugins: Ok this makes sense to me now.

pippinsplugins:
Opened new PR: https://github.com/restrictcontentpro/restrict-content-pro/pull/403

PR #346, opened 2016-02-09
update rcp_stripe_update_discount() to use args passed from action

bearded-avenger:
This function should utilize the args passed in from the action instead of $_POST fields.


bearded-avenger:
Note: ideally rcp_stripe_create_discount() should also be updated to use those args as well

pippinsplugins:
:+1:

pippinsplugins:
Thanks!

michaelbeil:
Oh great bearded one, we thank thee for thine gift.

Issue #347, opened 2016-02-09
2Checkout: RECURRING_INSTALLMENT_SUCCESS not processed


pippinsplugins:
When 2Checkout processes a renewal payment, it sends an INS with a message type of `RECURRING_INSTALLMENT_SUCCESS`.

Turns out we are not doing anything with this at the moment.

That means no renewal payments are getting logged.

Oops.

PR #350, opened 2016-02-10
Handle Recurring subscriptions with 2Checkout

tannerm:
#347 

Also did a little bit of clean up of some of the other hooks.

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-2checkout.php
+++ b/includes/gateways/class-rcp-payment-gateway-2checkout.php
@@ -211,10 +211,10 @@ class RCP_Payment_Gateway_2Checkout extends RCP_Payment_Gateway {
 					$payment_data = array(
 						'date'             => date( 'Y-m-d g:i:s', strtotime( $_POST['timestamp'], current_time( 'timestamp' ) ) ),
 						'subscription'     => $member->get_subscription_name(),
-						'payment_type'     => $payment_type,
+						'payment_type'     => sanitize_text_field( $_POST['payment_type'] ),
 						'subscription_key' => $subscription_key,
 						'amount'           => sanitize_text_field( $_POST['invoice_list_amount'] ),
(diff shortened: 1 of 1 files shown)


includes/gateways/class-rcp-payment-gateway-2checkout.php
```diff
@@ -241,6 +242,21 @@ public function process_webhooks() {
...
+						'subscription'     => $member->get_subscription_name(),
+						'payment_type'     => sanitize_text_field( $_POST['payment_type'] ),
+						'subscription_key' => $subscription_key,
+						'amount'           => sanitize_text_field( $_POST['item_list_amount_1'] ), // don't have a total from this call, but this should be safe
+						'user_id'          => $member->ID,
+						'transaction_id'   => sanitize_text_field( $_POST['invoice_id'] )
+					);
+
+					$recurring = ! empty( $_POST['recurring'] );
+					$member->renew( $recurring );
```
pippinsplugins: This line should be removed now since the renew() action is fired immediately after successful payment (line 154)

includes/gateways/class-rcp-payment-gateway-2checkout.php
```diff
@@ -222,7 +222,7 @@ public function process_webhooks() {
 
 					$recurring = ! empty( $_POST['recurring'] );
 					$member->renew( $recurring );
-					$member->add_note( __( 'Subscription cancelled via refund 2Checkout', 'rcp' ) );
+					$member->add_note( __( 'Subscription started in 2Checkout', 'rcp' ) );
```
pippinsplugins: This line should be removed now since the renew() action is fired immediately after successful payment (line 154)

pippinsplugins: The payment data also should not be created / inserted here. This webhook fires immediately after the payment is submitted so this will result in duplicate payments.

Issue #362, opened 2016-02-25
WP Approve User improperly disallows members when no restrictions on post


pippinsplugins:
If the WP Approve User plugin is activated and a non-approved (but registered) user attempts to view a page that has zero restrictions, they are shown a message that their account is pending verification.

When there are no restrictions on the page, the WP Approve User integration should not affect the page.

pippinsplugins:
@mindctrl @michaelbeil could one or both of you test this?

Let me know if you're unsure of how to replicate the bug.

mindctrl:
@pippinsplugins here are my test results.

Unrestricted page: content is visible - works as expected.

Restricted to a membership level: content not visible. pending msg shown. works as expected.

Restricted to paid only: content not visible. pending msg shown. works as expected.

Restricted to user level 1 (when NOT subscribed to a membership that gives level 1 access): content not visible. pending msg shown. works as expected.

Restricted to user level 1 (when IS subscribed to a membership that gives level 1 access): content is visible. not sure if this is expected or not since user is still pending.

Restricted to user level Subscriber: content is visible. not sure if this is expected or not since user is still pending.

pippinsplugins:
Does `user level 1` refer to a subscription level or an access level?

mindctrl:
the `Access Level` setting:

![image](https://cloud.githubusercontent.com/assets/1254870/13331752/ea7a83f0-dbcc-11e5-936e-3738e412b144.png)

pippinsplugins:
Thanks. That would not be the expected behavior so still a bug somewhere.

mindctrl:
@pippinsplugins appears that part of it is the `RCP_WP_Approve_User` class is instantiated too early and returns early in `__construct()` because the `Obenland_Wp_Approve_User` class isn't yet available. Did a quick test of firing it up in a function hooked on `plugins_loaded` and that prevents the early return, but it still didn't hide the content.

pippinsplugins:
@mindctrl Let's work on getting a PR sent to Obenland to resolve this.

mindctrl:
@pippinsplugins submitted a PR with an alternate approach that requires no upstream patch.

pippinsplugins:
@mindctrl Looks great.

mindctrl:
@michaelbeil mind giving this a quick test run before we close the issue? It's merged into `release/2.5` and needs the [WP Approve User plugin](https://wordpress.org/plugins/wp-approve-user/).

[...]

PR #363, opened 2016-02-25
Only show pending message to member when content is restricted. #362

pippinsplugins:
#362


mindctrl:
Closing since we went with https://github.com/restrictcontentpro/restrict-content-pro/pull/379

Issue #357, opened 2016-02-19
Stripe Gateway: Zero-decimal Currencies Are Not Calculated Correctly


michaeluno:
It seems by using the Stripe gateway [zero-decimal currencies](https://support.stripe.com/questions/which-zero-decimal-currencies-does-stripe-support) such as Japanese Yen get multiplied by 100. I guess it is the same problem reported on the EDD support forum [here](https://easydigitaldownloads.com/forums/topic/stripe-amounts-two-digits-off/).

mindctrl:
@michaeluno thanks for reporting this. We'll take a look at it for the next update.

pippinsplugins:
@mindctrl We've fixed this in EDD Stripe: 

https://github.com/easydigitaldownloads/edd-stripe/blob/master/edd-stripe-new.php#L1009

https://github.com/easydigitaldownloads/edd-stripe/blob/master/edd-stripe-new.php#L152

tannerm:
Just submitted a PR for this (#381).

PR #381, opened 2016-03-09
Issue/357

tannerm:
#357

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -606,3 +606,47 @@ function rcp_get_ip() {
 	}
 	return apply_filters( 'rcp_get_ip', $ip );
 }
+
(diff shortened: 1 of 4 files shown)


includes/misc-functions.php
```diff
@@ -647,4 +647,52 @@ function rcp_is_restricted_content( $post_id ) {
...
+
+/**
+ * Get RCP Currency
+ *
+ * @since 2.5
+ * @return mixed|void
+ */
+function rcp_get_currency() {
+	global $rcp_options;
+	return apply_filters( 'rcp_get_currency', $rcp_options['currency'] );
```
pippinsplugins: Let's add a default here in case the `currency` key is not set.

Issue #158, opened 2015-05-18
Get all User Roles for metabox's User Level setting


cliffordp:
I'm unclear if "User Level" is referring to WordPress' user roles, but I think it is.
If it is, then shouldn't it display all user roles, not just a hard-coded list of the default ones?

References:
- https://github.com/pippinsplugins/Restrict-Content-Pro/blob/master/includes/admin/metabox.php#L56
- http://codex.wordpress.org/Function_Reference/get_editable_roles

pippinsplugins:
It is default user roles. Custom roles aren't supported yet.

pippinsplugins:
Another request https://secure.helpscout.net/conversation/243086775/40810/?folderId=841776

mindctrl:
Started this in `issue/158`, but we need to consider what we'll do inside `includes/content-filters.php`.

pippinsplugins:
:+1:

nosegraze:
Can't we merge this as-is now that we got rid of that big `rcp_user_level_checks()` function in `content-filters.php` ?

nosegraze:
PR: #1156

pippinsplugins:
@nosegraze can you test this with a non-English site to ensure the translated role names work properly.

nosegraze:
@pippinsplugins Still works perfectly! The `value` in the metabox dropdown is actually more like a role ID so that doesn't get translated (the first parameter [here](https://codex.wordpress.org/Function_Reference/add_role)). They call it a name but really it's an ID.

pippinsplugins:
Excellent :+1:

PR #407, opened 2016-03-23
Update metabox.php

madvik:
it missing the roles created by development or plugins.


pippinsplugins:
@madvik We cannot show custom roles here yet because the functions that check whether a role has access to view content does not yet support custom roles.

See #158

PR #415, opened 2016-03-29
Bug/paypal upgrades

pippinsplugins:


mindctrl:
Seeing this during PayPal Standard purchases. `Undefined variable: include_fees in /html/wp-content/plugins/restrict-content-pro/includes/registration-functions.php on line 204`.

Members remain pending. No payment is logged in RCP. I see IPN requests in my logs.

PR #420, opened 2016-03-31
Prorate fix

jmarx:


jmarx:
Looks like this solution is causing some side-effect problems for me. Hang tight for a bit.

jmarx:
Yeah I have to take this back to the drawing board. The solution is not working for me :-1:

jmarx:
I _think_ that should do it.
The display was showing up right before but the actual invoices were still giving me the full price minus a dollar.
I am seeing a couple weird things on my local site, but I think that is because I have some custome stuff going on. Will keep testing. Probably done for today though. I hope you guys can get a crack at it.

pippinsplugins:
@jmarx Looks good now!

PR #461, opened 2016-04-18
Add currency signs to rcp_currency_filter

michaelbeil:
Fixes #456

PR #463, opened 2016-04-19
Issue/462

### LINKED ISSUE #462: Travis: in favor of less builds
I propose removing PHP nightly for the time being, unless we think it is mission critical. Also, we could remove test builds for 4.0-4.2 to speed up our build process.

michaelbeil:
Fixes #462


michaelbeil:
Unit tests pending #461 merge.

Issue #490, opened 2016-05-04
Update [restrict] to use can_access() method of RCP_Member


ghost1227:
The `[restrict]` shortcode currently doesn't use the `can_access()` method of `RCP_Member`. As a result, content restricted with the shortcode doesn't work with plugins like IP Restriction that override `can_access()`.

mindctrl:
PR attached. Needs thorough testing of the restrictions settings. During my tests, I found issues with restrictions under certain conditions, so went ahead and reworked the logic that handles that so the filter would return more accurate data. Need to make sure I didn't break something else.

pippinsplugins:
No more `extract()` 🎉

pippinsplugins:
@mindctrl This looks to work perfectly for everything except the `WP Approve User` integration. ANnunapproved user can still access restricted content.

pippinsplugins:
@mindctrl Could you test ^?

pippinsplugins:
Looks like subscription level IDs need to be trimmed.

Works: `[restrict subscription="1,2" paid="true"]`

Doesn't work: `[restrict subscription="1, 2 " paid="true"]`

pippinsplugins:
Level ID trimming fixed.

Just finished testing a bunch of configurations and everything seems to work great for me.

mindctrl:
@pippinsplugins WP Approve User integration looks great.

pippinsplugins:
Once unit tests pass, we'll merge.

michaelbeil:
Good deal!

PR #496, opened 2016-05-11
Introduce the rcp_restrict_shortcode_has_access filter to the [restrict] shortcode. #490.

mindctrl:
#490


includes/shortcodes.php
```diff
@@ -12,104 +12,104 @@
...
+	}
+
+	if ( $atts['userlevel'] === 'none' && ! is_user_logged_in() ) {
+		$has_access = false;
+	}
+
+	$has_access = (bool) apply_filters( 'rcp_restrict_shortcode_has_access', $has_access, $user_ID, $atts );
+
+	// admins always have access
+	if ( current_user_can( 'manage_options' ) ) {
```
pippinsplugins: Let's put this above the filter in the case someone wants to prevent admins from having access.

includes/shortcodes.php
```diff
@@ -12,104 +12,104 @@
...
-		} elseif ( $userlevel == 'subscriber' && current_user_can( 'read' ) && $has_access ) {
-		 	return apply_filters( 'rcp_restrict_shortcode_return', $content );
-		} elseif ( $userlevel == 'none' && is_user_logged_in() && $has_access ) {
-		 	return apply_filters( 'rcp_restrict_shortcode_return', $content );
-		} else {
-			return '<div class="rcp_restricted">' . rcp_format_teaser( $teaser ) . '</div>';
+		$classes = 'rcp_restricted';
+	}
+
+	if ( ! empty( $subscription ) && ! empty( $subscription[0] ) ) {
```
pippinsplugins: Why `'active' !== rcp_get_status( $user_ID )` and not `rcp_is_active()`?

includes/shortcodes.php
mindctrl: @pippinsplugins I originally had it that way, and moved it thinking people normally shouldn't/wouldn't restrict admins. I'll move it back up.

includes/shortcodes.php
mindctrl: Brain fart? That's my answer.

PR #506, opened 2016-05-19
Shortcodes - Removes the extract() function from all the shortcodes and support the "shortcode_atts_{$shortcode}" filter where attributes exists.

ramiy:


mindctrl:
@ramiy thanks! Can you not add line breaks after the arrays so that instead of this:
```
	$atts = shortcode_atts(
		array(
			'userlevel'    => 'none',
			'message'      => '',
			'paid'         => false,
			'level'        => 0,
			'subscription' => ''
		),
		$atts,
		'restrict'
	);
```

it still looks like this:

```
	$atts = shortcode_atts(
		array(
			'userlevel'    => 'none',
			'message'      => '',
			'paid'         => false,
			'level'        => 0,
			'subscription' => ''
		), $atts, 'restrict' );
```

ramiy:
@mindctrl Done.

mindctrl:
@ramiy thanks! Sorry I missed this. I'm going to log an issue and link it here so we don't miss it on the next release.

Issue #368, opened 2016-03-01
"Last Week" earnings report is incorrect


mindctrl:
The date range shown is incorrect, and there are some debug notices.

![image](https://cloud.githubusercontent.com/assets/1254870/13415688/c6bcc250-df2a-11e5-83ce-440f0a7c497b.png)

`[29-Feb-2016 21:13:49 America/New_York] PHP Notice:  A non well formed numeric value encountered in /Sites/rcp/wp-includes/functions.php on line 295`

![image](https://cloud.githubusercontent.com/assets/1254870/13415699/ec77afd2-df2a-11e5-8fba-8307548ded62.png)

michaelbeil:
Was just going to share this too and then thought I remembered there being an issue. I have just confirmed here that last week is not working.

pippinsplugins:
@michaelbeil @mindctrl mind testing this?

mindctrl:
@pippinsplugins this line seems to be an issue when `last_week` crosses a month. different bug but related and broken. https://github.com/restrictcontentpro/restrict-content-pro/blob/c26c4ac28d02e6dfbab33c792a85c2a7dd56e08a/includes/admin/reports/reports-page.php#L115

Dump of the `$dates` array:

![image](https://cloud.githubusercontent.com/assets/1254870/15836309/31b5a170-2c02-11e6-86d1-b0568e9657f8.png)

michaelbeil:
Still having an issue here.

pippinsplugins:
The dates array looks correct to me. The beginning of last week is 5/30 and the end of last week is 6/30.

pippinsplugins:
I see what it is. We accounted for this in EDD with this:
```php
} elseif ( $dates['range'] == 'this_week' || $dates['range'] == 'last_week' ) {

		$num_of_days = cal_days_in_month( CAL_GREGORIAN, $dates['m_start'], $dates['year'] );

		$report_dates = array();
		$i = 0;
		while ( $i <= 6 ) {

			if ( ( $dates['day'] + $i ) <= $num_of_days ) {
				$report_dates[ $i ] = array(
					'day'   => (string) $dates['day'] + $i,
					'month' => $dates['m_start'],
					'year'  => $dates['year'],
				);
			} else {
				$report_dates[ $i ] = array(
					'day'   => (string) $i,
					'month' => $dates['m_end'],
					'year'  => $dates['year_end'],
				);
			}

			$i++;
		}

[...]

PR #532, opened 2016-06-02
Fix incorrect date format passed to get_weekstartend(). #368

pippinsplugins:
#368


michaelbeil:
Closing in favor of #576, which is PRed against master.

Issue #544, opened 2016-06-09
Amount not properly sanitized in IPN verification


pippinsplugins:
The amount comparison done in the PayPal standard IPN fails sometimes because one of the amounts is not properly sanitized, which can result in one number having decimals and one not.

Example ticket where it happened: https://secure.helpscout.net/conversation/207731955/35077/?folderId=841781

pippinsplugins:
Going to merge this in and then put it on pippinsplugins.com

PR #545, opened 2016-06-09
Format pending amount. #544

pippinsplugins:
#544


michaelbeil:
Travis was failing, butI forced another build on 5.6 and all went well.

Issue #535, opened 2016-06-03
Revamp CSS 


michaelbeil:
There a few things we could do to better our CSS:
- [ ] Change px to rem
- [x] Remove empty selectors
- [x] Should our CSS worth be minified?
- [x] Make our CSS more readable
- [x] Replace super generic class names in admin with prefixed class names
- [x] Remove RCP-Icons since it's not used

pippinsplugins:
Just added another.

nosegraze:
Correct me if I'm wrong, but it looks like we're loading the `RCP-Icons` font in `forms.css` but then not actually using it anywhere?

mindctrl:
I think you're right, I noticed that a couple weeks ago. I vote we double check that, and get rid of it if so.

nosegraze:
I did a few searches and the only result I found for setting the `font-family` to `RCP-Icons` was here:

```css
[class*="rcp-icon-"], [class*=" rcp-icon-"] {
    font-family: 'RCP-Icons';
```

So I did a search for `rcp-icon` and there weren't any outside of the CSS file, so that class doesn't seem to be used anywhere.

mindctrl:
Confirmed. Added removing RCP-Icons to the list up top.

pippinsplugins:
:+1:

nosegraze:
WIP PR: #966

Did a code format on it so it looks like there are more changes than there are. :P

- Added some nice section and file headers.
- Tried to make spacing and bracket placement more consistent.
- Removed `@font-face` and associated classes from the CSS, though I haven't actually deleted the font files.
- Removed a small amount of duplicate code.
- Did some reorganizing to try to group similar items together.

Still needs more work and a more thorough look over.

pippinsplugins:
Looking  good so far 🎉

mindctrl:
Yay, looking good!

Nitpicky, but I'd prefer to keep tabs instead of moving to spaces. Just converting the existing files to tabs shaves over 600 bytes off `admin-styles.css` and around 300 bytes off `forms.css`... and tabs are consistent with our PHP and JS indentation.

pippinsplugins:
>>>Nitpicky, but I'd prefer to keep tabs instead of moving to spaces.

Tabs for indention please :)

nosegraze:
Updated with tabs!

PR #552, opened 2016-06-17
Remove fixed widths from labels + more

### LINKED ISSUE #251: Remove fixed widths from labels + more
We've discussed this before. Just realised I posted on PR. See: https://github.com/pippinsplugins/restrict-content-pro/pull/169#issuecomment-145329385

In summary:

1. Fixed widths on labels should be removed

2. The `Remember` checkbox (let's also rename this to Remember Me) doesn't look good on the 2015 or upcoming 2016 theme

3. Floats on labels should be removed (I see no need for them) 

I think we can fix this by either:

1. letting users know in advance about the changes
2. including the changes prominently in release notes
3. ???

michaelbeil:
Fixes #251

Assists with #535


pippinsplugins:
I don't see any inline widths removed.

michaelbeil:
I think it is better said, `fixed widths`, as you are right there were no inline widths removed, as we do not have any in our template files.

PR #568, opened 2016-06-30
Stripe Checkout i18n parameter

wpmarmite:
Language parameter is "locale" and no "local"


michaelbeil:
This breaks tests. Is there something not working on your site due to this parameter?

We would need to change other instances of local to account for this.

pippinsplugins:
@michaelbeil our tests are not passing right now. Not due to the PR.

Issue #592, opened 2016-07-19
Subscription name blank in emails for manually added users


mindctrl:
When adding a user manually, the `%subscription_name%` is empty in the email sent to the member. This is because we set the member status before setting the subscription level meta here: https://github.com/restrictcontentpro/restrict-content-pro/blob/2.6.3/includes/process-data.php#L122-L127

mindctrl:
@pippinsplugins this look good to you?

mindctrl:
Oh wait I see your comment on the PR.

PR #593, opened 2016-07-19
When manually adding users, set the subscription level before setting the status. #592

mindctrl:
#592

### DIFF
--- a/includes/process-data.php
+++ b/includes/process-data.php
@@ -117,14 +117,13 @@ function rcp_process_data() {
 					update_user_meta( $user->ID, '_rcp_new_subscription', '1' );
 				}
 
+				update_user_meta( $user->ID, 'rcp_signup_method', 'manual' );
+
+				update_user_meta( $user->ID, 'rcp_subscription_level', $level_id );
+
 				$status = $subscription->price == 0 ? 'free' : 'active';
 
 				rcp_set_status( $user->ID, $status );
-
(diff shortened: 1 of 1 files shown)


includes/process-data.php
```diff
@@ -117,15 +117,14 @@ function rcp_process_data() {
...
-				rcp_set_status( $user->ID, $status );
-
 				update_user_meta( $user->ID, 'rcp_signup_method', 'manual' );
 
-				// Add a role, if needed, to the user
 				update_user_meta( $user->ID, 'rcp_subscription_level', $level_id );
 
+				$status = $subscription->price == 0 ? 'free' : 'active';
+
+				rcp_set_status( $user->ID, $status );
```
pippinsplugins: While we're here, let's please go ahead and update this to use RCP_Member.

mindctrl: @pippinsplugins `rcp_set_status()` uses RCP_Member. Am I misunderstanding what you mean?

pippinsplugins: I meant use the `set_status()` method of RCP_Member so that we can slowly move away from all of the manual update_user_meta() calls in this file. We'd like to instantiate RCP_Member just once in all of the processing here. Using RCP_Member and functions like rcp_set_status() would result in two+ instances.

pippinsplugins:
Looks good.

PR #603, opened 2016-07-23
Added and translated 70 missing language strings after reloading POT

sanderdewijs:
Reloaded the POT file in Poedit and discovered more then 70 strings that were not yet translated in Dutch. Now it's 100% translated.


pippinsplugins:
Thanks!

Issue #639, opened 2016-08-10
Content improperly restricted for Free members when using Access Level restrictions


mindctrl:
Free members are unable to view content when content is restricted like via access levels in the _Restrict this content_ meta box.

![image](https://cloud.githubusercontent.com/assets/1254870/17570973/78b98e4a-5f1c-11e6-8589-4590615ec3d3.png)

This is because we set the `_is_paid` post meta flag when saving posts if there are any paid subscription levels on the site at all. [Reference 1](https://github.com/restrictcontentpro/restrict-content-pro/blob/2.6.6/includes/admin/metabox.php#L146-L154) and [Reference 2](https://github.com/restrictcontentpro/restrict-content-pro/blob/2.6.6/includes/admin/metabox.php#L193)

We discussed getting away from the metadata at some point in the future. A possible workaround until then is to adjust the `can_access()` method in `RCP_Member` to check for expired instead of not is active. Changing [this](https://github.com/restrictcontentpro/restrict-content-pro/blob/2.6.6/includes/class-rcp-member.php#L736):

from:

`if ( rcp_is_paid_content( $post_id ) && ! $this->is_active() ) {`

to this:

`if ( rcp_is_paid_content( $post_id ) && $this->is_expired() ) {`

mindctrl:
That appears to work in my local testing. This doesn't appear to break anything because the paid level checks happen below that line. Going to push up a change and see what the unit tests think. My local phpunit is broken at the moment.

pippinsplugins:
:+1:

mindctrl:
@michaelbeil do you have a few minutes you can spare to test this?

mindctrl:
@pippinsplugins I think we're good here, but can you give it a whirl to make sure?

pippinsplugins:
Works properly for me.

PR #640, opened 2016-08-10
Change the access check in can_access in RCP_Members. #639.

mindctrl:
#639

### DIFF
--- a/includes/class-rcp-member.php
+++ b/includes/class-rcp-member.php
@@ -733,7 +733,7 @@ class RCP_Member extends WP_User {
 		// Assume the user can until proven false
 		$ret = true;
 
-		if ( rcp_is_paid_content( $post_id ) && ! $this->is_active() ) {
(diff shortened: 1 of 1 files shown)


includes/class-rcp-member.php
```diff
@@ -733,7 +733,7 @@ public function can_access( $post_id = 0 ) {
 		// Assume the user can until proven false
 		$ret = true;
 
-		if ( rcp_is_paid_content( $post_id ) && ! $this->is_active() ) {
+		if ( rcp_is_paid_content( $post_id ) && $this->is_expired() ) {
```
pippinsplugins: Wouldn't this improperly permit a non-paid, non-expired member to access paid content?

mindctrl: @pippinsplugins it's a bit confusing, but the paid checks happen below this line and check against is_active(), which prevents free members from accessing paid content.

pippinsplugins: :+1:

PR #641, opened 2016-08-11
Invoice improvements

ericnormand:
Two changes:

1. The header logo on my invoices was looking too big and getting cut off. This adds a CSS directive to all `img` tags to keep them within the bounds.

2. Add hooks for customizing the invoice information displayed. I personally use these to add information such as the member's business name and address, VAT ID, and VAT %, all of which are required for Europe.

### DIFF
--- a/templates/invoice.php
+++ b/templates/invoice.php
@@ -223,7 +223,7 @@
 					<?php if ( ! empty( $rcp_options['invoice_email'] ) ) : ?>
 						<p><strong><?php echo $rcp_options['invoice_email']; ?></strong></p>
 					<?php endif; ?>
-
+					<?php do_action( 'rcp_invoice_extra' ); ?>
 				</article>
 			</div>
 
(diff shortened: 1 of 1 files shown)


michaelbeil:
This looks good. We will check it out.

mindctrl:
@ericnormand thanks for the PR.

I pulled this down and started playing with it. At first I loved this idea, but after looking at the naming and positioning of the hooks, and considering that adding them will potentially lock us into the invoice structure we have now, I'd rather not add the hooks part of this PR.

The HTML template can be overridden in a theme for one-off uses. I'm sure as time goes on we'll have extensions that tie into this, but until then let's hold off on adding hooks. When we get there, I want the hooks to have more descriptive names.

Would you mind sending a new PR with only the CSS change?

ericnormand:
Sure. I'll do that.

How can I override the template with a theme?

Thanks
Eric

pippinsplugins:
See http://docs.pippinsplugins.com/article/656-template-files

PR #666, opened 2016-08-30
Make sure logo doesn't get too big on invoice

ericnormand:
The logo on my site was spilling out of the `div`. This makes sure it never gets bigger than the `div` it is contained in.


mindctrl:
Works great. Thanks @ericnormand. Merging. #667

PR #678, opened 2016-09-09
Remove unused globals

ramiy:
Remove unused `$wp_version` and `$rcp_load_css` globals from the code.

And replace `$current_user` global with `wp_get_current_user()` function ([read this](https://generatewp.com/quick-tip-stop-using-get_currentuserinfo/)).


mindctrl:
@ramiy we need to keep the `$rcp_load_css` global here for now. Removing it prevents the CSS from loading on the form if someone loads it programmatically with `rcp_get_template_part()`.

templates/profile-editor.php
```diff
@@ -2,8 +2,7 @@
 /**
  * This template is used to display the profile editor with [rcp_profile_editor]
  */
-global $current_user, $rcp_load_css;
```
mindctrl: We need to keep `$rcp_load_css`

ramiy: @mindctrl reverted.

PR #680, opened 2016-09-10
Apply a filter to the restrict option dropdown

joshuadavidnelson:
I'm hoping to expand these "restrict by" options for a client (working on a "restrict by group" add-on to work with the Group Account extension), but I figure this will be handy for others as well. There's probably a better way to build this out, and this doesn't address the javascript related to the viewing of the sub options. Appreciate consideration/input.

### DIFF
--- a/includes/admin/metabox-view.php
+++ b/includes/admin/metabox-view.php
@@ -20,10 +20,37 @@ $role_set_display  = '' != $user_role ? '' : ' style="display:none;"';
 	</p>
 	<p>
 		<select id="rcp-restrict-by" name="rcp_restrict_by">
-			<option value="unrestricted" <?php selected( true, ( empty( $sub_levels ) && empty( $access_level ) && empty( $is_paid ) ) ); ?>><?php _e( 'Everyone', 'rcp' ); ?></option>
-			<option value="subscription-level"<?php selected( true, ! empty( $sub_levels ) || ! empty( $is_paid ) ); ?>><?php _e( 'Members of subscription level(s)', 'rcp' ); ?></option>
-			<option value="access-level"<?php selected( true, is_numeric( $access_level ) ); ?>><?php _e( 'Members with an access level', 'rcp' ); ?></option>
-			<option value="registered-users"<?php selected( true, empty( $sub_levels ) && ! is_numeric( $access_level ) && ! empty( $user_role ) && 'All' !== $user_role ); ?>><?php _e( 'Members with a certain role', 'rcp' ); ?></option>
+		<?php
+			$array = array(
+				'unrestricted' => array(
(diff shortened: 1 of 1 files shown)


pippinsplugins:
Great idea for sure. My one concern, however, is that in order to really use custom options here (at least in an intuitive way) there also needs to be some custom JS written to hide/show appropriate elements.

Any thoughts or concerns with that?

joshuadavidnelson:
I completely agree. I looked at the JavaScript associated with this, but my skills there are a little weaker, so I thought I'd start here and test the waters before diving in further.

My approach would be to associate the dropdown value to a wrapper div with the custom options. So, you'd extend it by using the after hook to add associated custom fields wrapped by a div with an appropriate id. Looks like the javascript is written to reference the specific elements and dropdown values, so I'd update it to be more universal. Any other way you'd rather I do it?

mindctrl:
I like the idea, but I have a couple concerns.

First, because we just changed this UI in version 2.6 and had to make a couple bug fixes afterwards, I'm concerned we may introduce a bug here or otherwise make it more brittle. I'd rather this bake for while before making it more flexible.

Second, with several of our own add-ons adjusting content visibility, adding more flexibility adds more complexity and things we may need to consider, maintain, and test. It may end up not being an issue, but something we need to consider.

For JS, we should probably add some triggers to our existing JS to make that part easier to tie into.

With regards to 'restrict by group' functionality, that's a cool idea. I think it's a good idea for the Group Accounts add-on itself, as opposed to another add-on. @joshuadavidnelson would you be interested in contributing the feature to the Group Accounts add-on directly?

joshuadavidnelson:
John, all good points - I had a feeling such a change would bring up things like this.

In regards to contributing to the Group Accounts add-on - Yes, absolutely! I didn't see a repo for it, so I took a stab here to get the ball rolling. I agree it would make more sense to be part of that add-on than an add-on-to-an-add-on. ;-)

Let me know the best way for me to proceed if we go that route. Appreciate the consideration and flexibility!

mindctrl:
@joshuadavidnelson great! Sent you an invitation to the Group Accounts repo.

joshuadavidnelson:
@mindctrl Awesome, thanks! I'll issue a PR there sometime later this week, likely. In regards to the UI, do we shelf the idea of extending this dropdown or is there another approach I should use?

mindctrl:
@joshuadavidnelson let's leave this open for now. I'll think about this a bit and see what I come up with. Feel free to shoot me a message at restrictcontentpro.com/support if you'd like to chat directly.

joshuadavidnelson:
@mindctrl Sounds good. Thanks!

Issue #690, opened 2016-09-26
Use rcp_get_currency() throughout the plugin in place of direct options call


mindctrl:
`rcp_currency_filter()` should be calling `rcp_get_currency()` [here](https://github.com/restrictcontentpro/restrict-content-pro/blob/2.6.7/includes/misc-functions.php#L73
), instead of getting it directly from the options. This will allow the filter in `rcp_get_currency()` to work with `rcp_currency_filter()`.

mindctrl:
I see a couple other places we could probably do this.

mindctrl:
@pippinsplugins I updated this PR since you approved it. Mind giving it another look?

pippinsplugins:
PR looks good still.

PR #691, opened 2016-09-26
Use rcp_get_currency() in rcp_currency_filter() instead of getting the currency from options. #690

mindctrl:
#690


mindctrl:
Looks good here. Unit tests failed, only one two sets, due to connection timeout to wp.org.

`could not connect to server (https://develop.svn.wordpress.org)`

Issue #695, opened 2016-09-30
Can't complete registration with 100% discount with Stripe Checkout


mindctrl:
With Stripe Checkout enabled, if someone goes to the registration form and applies a 100% discount, the registration won't complete. It just sits there.

[Customer reported](https://secure.helpscout.net/conversation/257672347/43554/?folderId=841781) and confirmed locally.

pippinsplugins:
I just tested this and it worked properly for me.

mindctrl:
Hmm doesn't work for me on two different sites.

pippinsplugins:
MaIn registration form with all levels shown or just a single level?

mindctrl:
Main reg form with all levels shown.

mindctrl:
@pippinsplugins try that.

pippinsplugins:
@mindctrl works for me!

PR #700, opened 2016-10-03
Return true instead of false if 100% discounted. #695

mindctrl:
#695


mindctrl:
Merging this. Two unit tests failed due to SSL handshake with wporg. Everything else passed.

`svn: OPTIONS of 'https://develop.svn.wordpress.org/tags/4.4.2/tests/phpunit/includes': SSL handshake failed: Secure connection truncated (https://develop.svn.wordpress.org)`

PR #711, opened 2016-10-10
Add $discount, $subscription_id to rcp_discounts_once_per_user filter

machouinard:
Client has multiple discount codes with only one requiring single use limit.  

This provides more flexibility with the filter.


pippinsplugins:
Good idea!

Issue #595, opened 2016-07-19
Stripe subscription expiration dates being incorrectly extended


mindctrl:
Customer reported, and I've confirmed in tests, that sometimes a member's expiration date is extended an additional period. This happens when Stripe sends a webhook updating the status of the $0 invoice we create to set the currency type.

I tested this 3 different times, and it happened on 2 of them. It happens almost exactly an hour after signup.

Event log listing
![image](https://cloud.githubusercontent.com/assets/1254870/16959258/314d5e22-4db2-11e6-8fc3-234a08475ae5.png)

Note the top two entries are updates that happen automatically an hour later, initiated by Stripe.

![image](https://cloud.githubusercontent.com/assets/1254870/16959275/467743da-4db2-11e6-9b39-e61b2cea8ba2.png)

and

![image](https://cloud.githubusercontent.com/assets/1254870/16959281/4fce1bde-4db2-11e6-895f-e8b520934bd7.png)

Looking at the details, it seems like the `closed` status was not set on the 2 that this happened on, but the one that it did not happen on showed the $0 invoice was successfully closed within seconds of its creation, instead of an hour later.

@pippinsplugins thoughts on this?

pippinsplugins:
Ok so it's happening because the 0.00 invoice is firing a `invoice.payment_succeeded` event.

We could either forcibly delete that invoice after the customer has been deleted or perhaps find someway to detect pseudo invoices like that and not trigger a renewal process. A small piece of metadata on the invoice would probably suffice if we are not able to just delete the invoice.

mindctrl:
It looks like we only close the temp invoice if there's a signup fee, which would explain why it only happens sometimes. Reference: https://github.com/restrictcontentpro/restrict-content-pro/blob/2.6.3/includes/gateways/class-rcp-payment-gateway-stripe.php#L151-L162

I see in https://github.com/restrictcontentpro/restrict-content-pro/issues/549 that you mentioned that setting the currency was related to signup fees. Since we're doing the temp invoice regardless of signup fees, I propose we either only create the temp invoice when there are signup fees or move the invoice closing logic outside the conditional that checks for signup fees. @pippinsplugins

pippinsplugins:
Let's only create it if there are signup fees.

PR #719, opened 2016-10-24
Close stripe invoice regardless of renew status

JimChestnutt:
Current stripe gateway only attempts to close invoice if auto_renew is true. This change also places the invoice closing code into the single payment flow.

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-stripe.php
+++ b/includes/gateways/class-rcp-payment-gateway-stripe.php
@@ -249,6 +249,14 @@ class RCP_Payment_Gateway_Stripe extends RCP_Payment_Gateway {
 
 			try {
 
+				// Attempt to close invoice even if auto_renew is not set
+                if( isset( $temp_invoice ) ) {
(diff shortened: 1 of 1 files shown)


includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -249,6 +249,14 @@ public function process_signup() {
...
 			try {
 
+				// Attempt to close invoice even if auto_renew is not set
+                if( isset( $temp_invoice ) ) {
+                    $invoice = \Stripe\Invoice::retrieve( $temp_invoice->id );
+                    $invoice->closed = true;
+                    $invoice->save();
+                    unset( $temp_invoice, $invoice );
+                }
+
```
mindctrl: @JimChestnutt Needs to be indented with tabs. Also I think instead of duplicating the code we should rework it to run outside the conditionals.

mindctrl:
Testing this to ensure no unintended consequences related to https://github.com/restrictcontentpro/restrict-content-pro/issues/595.

Issue #708, opened 2016-10-06
Expiration date for pending members


mindctrl:
If you view the pending member list in admin, many/most of the members will show an expiration date of today. That's because they don't have a real expiration date, only a pending one. 

`rcp_get_expiration_date()` specifically calls the `get_expiration_date()` method in RCP_Member with the second param as false, which tells it not to use the pending expiration... which means it defaults to today.

After chatting about it, we decided to show the pending expiration date.

PR #727, opened 2016-10-26
Showing pending expiration date for pending members. #708

mindctrl:
#708


mindctrl:
Tested only failed due to connection issue with wordpress.org.

`svn: OPTIONS of 'https://develop.svn.wordpress.org/tags/4.6.1/tests/phpunit/includes': SSL handshake failed: Secure connection truncated `

Issue #709, opened 2016-10-07
Comments visible when post only restricted by role


mindctrl:
If a post is restricted only via user access/role, the comments on the post are visible if a user doesn't have access.

In `rcp_hide_comments()` we check for access with `rcp_user_can_access()`, but that doesn't check the role at all. 

I did a quick test by adjusting the `can_access()` method in `RCP_Member` to check for a role, and it seems to work ok, but we'd need to do some thorough testing if we go that route.

I did a quick test with the following:

Get the saved user level with this
`$user_level          = get_post_meta( $post_id, 'rcp_user_level', true );`

and then added this check
```
		if ( $ret && ! empty( $user_level ) ) {
			if ( ! user_can( $this->ID, strtolower( $user_level ) ) || $this->expired() ) {
				$ret = false;
			}
		}
```

pippinsplugins:
Seems like a good fix.

PR #730, opened 2016-10-26
Add role check to can_access() method in RCP_Member. #709

mindctrl:
#709


mindctrl:
All tests passed except one that failed due to SSL handshake error with wordpress.org.

`SSL handshake failed: Secure connection truncated `

Issue #354, opened 2016-02-13
Change password form is not validated 


thomasfw:
Users are able to leave the password fields blank when updating their password with `rcp_change_password_form()`. Can we include a validation check here to ensure that they're entering something (edit: I can also enter two different passwords, with no validation at all)?

michaelbeil:
Yes, we should work this into a future release.

pippinsplugins:
We'll get that fixed!

thomasfw:
Any reason this keeps getting pushed back? It's a fairly large bug imo?

pippinsplugins:
@thomasfw No reason other than urgency of other issues that had to be included in a release. This will get fixed.

pippinsplugins:
Validation is working fine for me:
![screen shot 2016-05-31 at 1 30 20 pm](https://cloud.githubusercontent.com/assets/1034109/15685989/dcecfe5e-2733-11e6-8ca5-d53c26396c28.png)
![screen shot 2016-05-31 at 1 30 33 pm](https://cloud.githubusercontent.com/assets/1034109/15685999/e46fab22-2733-11e6-830a-80ce83bf4860.png)

thomasfw:
Nope, this doesn't work for me, I'm using the `rcp_login_form_fields()` function in a custom template (and the result is the same if I output the `[login_form]` shortcode).

I visit the reset link as received by email..

`www.site.dev/login/?rcp_action=lostpassword_reset&key=[KEY]&login=[username]`

`rcp_login_form_fields()` function correctly displays `rcp_change_password_form()` (as rcp_action == 'lostpassword_reset'). This contains the fields for resetting the password, and the following hidden input..

`<input type="hidden" name="rcp_action" value="reset-password">`

So I submit the form with an invalid password combination, and rcp_action is changed from 'lostpassword_reset' to 'reset-password', meaning that when I'm redirected back to the login page, all I see is the default login form (and no error messages).

So the conditional in `rcp_login_form_fields()` needs to also check for 'reset-password'...

```
elseif ( isset($_REQUEST['rcp_action']) && ($_REQUEST['rcp_action'] === "lostpassword_reset" || $_REQUEST['rcp_action'] === "reset-password") ) {
		echo rcp_change_password_form();
```

This fixes the issue for me..

pippinsplugins:
Your original issue only mentioned changing the password, not resetting it. I'll test again to see if I can replicate.

geoffreybr:
[...]

PR #740, opened 2016-11-01
Fix not showing again reinit password form when error in this form. #354

geoffreybr:
When the user clicks on the link of the login form to reset his password then clicks on the link in the email he received to set a new password, he reaches the reinit password form. 
If everything is fine when he enters his new password, he is redirected properly to the login form and can log in. However, if he put nothing in the reinit password form or if the two password fields do not match, the same occurs : he is redirected to the login form (without error message !) whereas the password was not changed. So I receive complaints from people who think they have changed their password (whereas they mistook in the form) but cannot login because the password was not actually reset...

This fix from @thomasfw in #354 changes the behavior if the user mistakes : the reinit password form is shown again with the proper error message so the user knows he mistook and has to enter again the new password.


mindctrl:
Thanks @geoffreybr.

PR #755, opened 2016-11-07
Fix fatal error on unsupported PHP versions

nosegraze:
If someone's PHP version is lower than 5.3, none of the plugin files get loaded, but the `rcp_register_databases` action still triggers. This causes a fatal error when trying to use classes (like `RCP_Payments`) that don't exist.

Moving the action into the `else` block for the version check fixes the problem.


mindctrl:
@nosegraze good catch. Could you create an issue for this for easier tracking and milestone it to 2.6.11.

nosegraze:
Done: #756

Issue #483, opened 2016-05-02
Discounts: add support for one-time discounts


michaelbeil:
The current behavior of discounts applies a discount on every recurring payment of a subscription. We should add an option to apply a discount on solely the first payment. 

By default, recurring discounts will be enabled, but this option could be enabled from the Misc tab to disable recurring discounts and enable first payment discounts.

derickbailey:
i'm very much looking forward to this one. right now i have to create a new plan with a negative signup fee, but make everything else match the original plan. having one-time discounts will greatly improve things for me!

pippinsplugins:
I have this started and working for PayPal Standard.

Adding support for the other gateways is a bit trickier. For PayPal Pro and Express, we'll likely need to completely refactor those gateways to mimic the ones in EDD Recurring Payments.

Todo:

- [ ] Stripe
- [ ] PayPal Express
- [ ] PayPal Pro
- [ ] 2Checkout

pippinsplugins:
Needs testing:

- [ ] PayPal Standard
- [ ] 2Checkout
- [ ] PayPal Pro
- [ ] PayPal Express
- [ ] Stripe

nosegraze:
A bit late, I suppose, but I just got a request for this, but also for an option to specify how many times the discount code recurs for (i.e. it applies for 3 months, then returns to full amount). Would be cool. :P

pippinsplugins:
That would be sweet.

cklosowski:
@nosegraze I have something sort of working for that with EDD Recurring and stripe. @pippinsplugins where do we stand on this? Do we need a few gateways updated still or is it basically ready to go? The request @nosegraze got was from someone I know that's looking for this and I'd be happy to jump in and help out.

pippinsplugins:
@cklosowski I still have more work to do on it. I've got it mostly working in 2checkout and paypal standard.

pippinsplugins:
Ok one-time discounts are working for Stripe!

Note: if one-time discounts are enabled, we do _not_ use Stripe coupons. Instead we use the customer balance feature. If one-time discounts are disabled, we use the Stripe coupons.

derickbailey:
SUPER happy to see this work done! this has been a big thing on my list of needs for a while now. :)

i'm curious, though - what's the reason behind the decision for using / not using stripe coupons? stripe supports one-time use discount codes... i'm not familiar with how the "customer balance feature" works, so i'm probably missing something here.

[...]

PR #758, opened 2016-11-08
WIP. First pass at adding support for one-time discount codes. issue/483

pippinsplugins:
#483


includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -192,7 +207,7 @@ public function process_signup() {
 					'prorate' => false
 				);
 
-				if ( ! empty( $this->discount_code ) ) {
+				if ( ! empty( $this->discount_code ) && ! isset( $rcp_options['one_time_discounts'] ) ) {
```
nosegraze: The `$rcp_options` global needs to be defined before being used here. Otherwise this is always true.

Issue #497, opened 2016-05-13
Add restriction settings to the Add term screen


mindctrl:
Right now the term restriction settings are only visible on the `edit term` screen. This results in people not even knowing it's a feature. Let's see about adding them to the `add term` screen.

pippinsplugins:
Moving to 2.7.

pippinsplugins:
Started in #760

pippinsplugins:
@mindctrl I think this is ready to merge, unless you want to play with the UI some more.

pippinsplugins:
Tested this with my Sugar Event Calendar plugin and the category restriction worked perfectly for it's custom taxonomy.

mindctrl:
I'll go ahead and merge it. The UI issue I see is really a core thing, due to the fact that the add term happens via ajax, so if the list is long enough to push it out of view, I don't think there's much we can do without some hacky javascript. If I come up with an idea I'll work it from another issue.

PR #760, opened 2016-11-08
Add restriction options to add taxonomy term form - issue/497

pippinsplugins:
WIP.

Also renames functions / file to be generic instead of for categories. #497

### DIFF
--- a/includes/admin/terms.php
+++ b/includes/admin/terms.php
@@ -91,41 +91,36 @@ function rcp_term_edit_meta_fields( $term ) {
  */
 function rcp_term_add_meta_fields( $taxonomy ) {
 	?>
-	<tr>
-		<th scope="row"><?php _e( 'Paid Only?', 'rcp' ); ?></th>
-		<td>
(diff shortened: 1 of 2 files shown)


mindctrl:
@pippinsplugins anything else you wanted to do here?

pippinsplugins:
@mindctrl I don't believe it's 100% working yet. Need to relook at it.

mindctrl:
Updated this to hook into `created_term` to save on new term creation. One thing I noticed about the UX, which is really a WP core thing I guess, is if you have enough levels to push the top of the screen out of view, when you click `Add New Category` you can't tell that the term was added. We may need to figure out a workaround for this.

mindctrl:
Circled back around to this with some fresher eyes. I removed the table markup here, since there's no table we're inserting into already. I matched the core convention with `<div>` wrappers around each "section". That made things cleaner looking.

pippinsplugins:
This works great.

Issue #351, opened 2016-02-10
Integration: Easy Digital Downloads


mindctrl:
Having tighter EDD integration would be great.

Some areas to consider:
- Purchase link (currently not protected on themes that use it manually - in a sidebar for example)
- Prevent checkouts if someone gets an item in the cart by a direct link
- Download links (prevent inactive subscribers from accessing downloads)
- Login redirect for those without active subscriptions when using [edd_login]
- Revoke EDD API access?

Could be others I'm not thinking about now. We can create a formal list after more consideration. Please add any others you think of.

pippinsplugins:
I'd add: restricting `[downloads]` display and downloads archive to paid subscribers.

pippinsplugins:
Instead of some of the options we included above, let's do this just like the WooCommerce integration we have.

mindctrl:
Started mocking this up in `issue/351`. Will continue building it out.

mindctrl:
I don't remember what I meant here: `Login redirect for those without active subscriptions when using [edd_login]`

mindctrl:
> let's do this just like the WooCommerce integration we have

@pippinsplugins our Woo integration has separate settings for viewing and purchasing, but the UI is totally different. On EDD products we use the standard restriction UI. Adding another set of those same controls for controlling purchasing separate will not look good in my opinion. Right now the code on `issue/351` just ties purchasing ability to the viewing. Do you want us to separate the two and explore UI options?

mindctrl:
I'm ok with doing it like it is now and splitting them later if we get enough requests.

pippinsplugins:
I'm fine with just limiting viewing / purchasing with a single UI.

mindctrl:
Added a PR to make it easier to see. We can continue working on it.

mindctrl:
Checklist of items tested:

* [x] Download links do not work if member is expired (edd_file_download_has_access filter)
* [x] Removing restricted items from the /downloads archive listing, if "Hide Premium Posts" is enabled and the member doesn't have access.
* [x] Removing restricted items from [downloads] shortcode output if "Hide Premium Posts" is enabled and the member doesn't have access.
* [x] Redirecting user if "Hide Premium Posts" is enabled, a redirect page is set, and the user doesn't have access to the single download.
* [x] Remove Add to Cart button if member doesn't have access to product.
[...]

PR #764, opened 2016-11-11
Add Easy Digital Downloads integration. #351

mindctrl:
#351 

Leaving unmerged until we firm everything up.

### DIFF
--- a/includes/integrations/class-rcp-easy-digital-downloads.php
+++ b/includes/integrations/class-rcp-easy-digital-downloads.php
@@ -13,6 +13,7 @@ class RCP_EDD {
 		add_filter( 'edd_purchase_download_form', array( $this, 'download_form' ), 10, 2 );
 		add_filter( 'edd_file_download_has_access', array( $this, 'file_download_has_access' ), 10, 3 );
 		add_filter( 'edd_downloads_query', array( $this, 'edd_downloads_query' ), 10, 2 );
+		add_filter( 'edd_downloads_excerpt', array( $this, 'edd_downloads_excerpt' ) );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/integrations/class-rcp-easy-digital-downloads.php
```diff
@@ -0,0 +1,92 @@
...
+
+
+function rcp_edd_init() {
+
+	if ( ! class_exists( 'Easy_Digital_Downloads' ) ) {
+		return;
+	}
+	new RCP_EDD;
+}
+add_action( 'init', 'rcp_edd_init' );
```
pippinsplugins: Some of the EDD used above are tied to `init`. Please double check the priorities to ensure using default `10` works.

mindctrl: Everything worked in my tests, but I'll be doing some more thorough testing over the next couple days... with different extensions, different servers, etc.

includes/integrations/class-rcp-easy-digital-downloads.php
```diff
@@ -79,6 +80,35 @@ public function edd_downloads_query( $query, $atts ) {
...
+	 * @access public
+	 * @since 2.7
+	 */
+	public function edd_downloads_excerpt( $excerpt ) {
+
+		global $rcp_options;
+
+		$post_id = get_the_ID();
+
+		if ( $this->member->can_access( $post_id ) ) {
```
pippinsplugins: This should show the excerpt if `Show excerpt` is enabled on the product.

mindctrl: @pippinsplugins good catch. updated.

Issue #769, opened 2016-11-16
Add filters and actions in rcp_process_profile_editor_updates


geoffreybr:
I added few fields in the profile editor ('Edit your profile' page) : URL of the user's Facebook page, a small description of the user, etc. ; the goal was to show these contents in the comments of the registered users without giving them access to the WP back-office.
In `rcp_process_profile_editor_updates`, there is no filters and actions equivalent to `rcp_form_errors` and `rcp_form_processing` that are used in the registration form. So I had the choice between modifying the code of restrict content pro (which is not a good idea for updates) and replacing the `rcp_process_profile_editor_updates` action by mine (which I did).
Moreover, it could be great to add the old user data in the `rcp_user_profile_updated` action. For example, I use the old email address to check if it has changed and, if so, I send a verification link (#632).

It would be great to add these kind of actions and filters to allow the plug-ins to really plug in the profile editor :slightly_smiling_face:. I will also use them to improve the Mailchimp plug-in.

*Note: As usual, the pull request is coming*

pippinsplugins:
This looks good to me!

mindctrl:
Merged. Thanks @geoffreybr

geoffreybr:
@mindctrl Do you know when the v2.7 will be released ? I will need this soon (so will probably have to integrate the feature in my version) but it could be good to have an approximative date :wink:. Thanks!

mindctrl:
@geoffreybr we don't have a set date yet. My guess would be within the next 2 months, but that depends largely on how things go during the holiday season.

PR #770, opened 2016-11-16
Add filters and actions in rcp_process_profile_editor_updates

geoffreybr:
See #769 for details.

Here is an example to add a mandatory 'Description' field to the 'Edit your profile' page: 
https://gist.github.com/geoffreybr/d2dcd2d8db96c432f6f1866099b81d09

### DIFF
--- a/includes/member-functions.php
+++ b/includes/member-functions.php
@@ -798,6 +798,8 @@ function rcp_process_profile_editor_updates() {
 		}
 	}
 
+	do_action( 'rcp_edit_profile_form_errors', $_POST, $user_id );
+
 	// retrieve all error messages, if any
 	$errors = rcp_errors()->get_error_messages();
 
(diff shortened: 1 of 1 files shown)


includes/member-functions.php
```diff
@@ -798,6 +798,8 @@ function rcp_process_profile_editor_updates() {
 		}
 	}
 
+	do_action( 'rcp_edit_profile_form_errors', $_POST, $user_id );
```
mindctrl: @geoffreybr I think it might be useful to also pass `$old_data` to this hook too, like you did with the one below.

mindctrl: @geoffreybr Retrieving it again requires duplicating the query. Originally I was thinking what if I want to compare old data to new posted data before updating? I think it's fine either way.

mindctrl: @geoffreybr `$userdata` is made up of `$_POST` data, which you already have, and `$user_id`. Let's just leave it like you have it for now. If people start asking for old data, we can add it in the future.

Issue #492, opened 2016-05-10
Make [subscription_details] responsive


madvik:
By default, the table `#rcp-account-overview` and `#rcp-payment-history` in `subscription.php` was not responsive.
I propose to add  `data-th="Title"` in `<td>` for transform in responsive table like this : [https://codepen.io/geoffyuen/pen/FCBEg](https://codepen.io/geoffyuen/pen/FCBEg)

mindctrl:
@madvik thanks for the suggestion. That's an interesting solution. We'll take a look at improving the responsiveness of the tables.

mindctrl:
@madvik looking at your PR, this is what I see

![image](https://cloud.githubusercontent.com/assets/1254870/20526903/1dd1804a-b094-11e6-87f7-5dbac66efb55.png)

---

![image](https://cloud.githubusercontent.com/assets/1254870/20526875/060eac3a-b094-11e6-818e-be72238cee64.png)

Are you seeing the same thing? If so, we just need some refinements at breakpoints where content is running together or wrapping oddly. Definitely an improvement and move in the right direction.

mindctrl:
@madvik nice improvement!

Mobile view:
![image](https://cloud.githubusercontent.com/assets/1254870/20535269/283b7cc0-b0b2-11e6-8090-048d622fd64f.png)

![image](https://cloud.githubusercontent.com/assets/1254870/20535311/4dc16a4a-b0b2-11e6-88b0-27227ca7efb4.png)

I'll give this another code review later today (lunch calls) and will merge if all looks well.

madvik:
Thanks a lot

helsinglee:
@mindctrl is it mean that I dont need to style the tables, currently the tables and pages were unstyled, that is cruel.

pippinsplugins:
This works perfectly. Merging and closing.

PR #778, opened 2016-11-18
Add attribute for responsive

madvik:
Proposition following to #492


mindctrl:
@madvik thanks for this PR and the other. Please combine this into the other one so it's all in one place.

Issue #783, opened 2016-11-23
Add support for user-defined locales


ramiabraham:
WordPress 4.7 introduces `get_user_locale()` ([changeset](https://core.trac.wordpress.org/changeset/38705)), which falls back to the standard `get_locale()`.

This changes the RCP `load_textdomain()` loader method ([line link](https://github.com/restrictcontentpro/restrict-content-pro/blob/master/restrict-content-pro.php#L131)) to support user locales.

ramiabraham:
Added a PR: https://github.com/restrictcontentpro/restrict-content-pro/pull/784

mindctrl:
Was going to test this, but don't see an option in the profile editor to change user language on the latest trunk. Will swing back around.

pippinsplugins:
Works perfectly!

PR #784, opened 2016-11-23
issue/783: Add support for user locales

ramiabraham:
- Adjusts language loader to provide support for `get_user_locale()`

#783

### DIFF
--- a/restrict-content-pro.php
+++ b/restrict-content-pro.php
@@ -128,7 +128,21 @@ function rcp_load_textdomain() {
 
 
 	// Traditional WordPress plugin locale filter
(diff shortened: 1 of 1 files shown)


restrict-content-pro.php
```diff
@@ -128,7 +128,21 @@ function rcp_load_textdomain() {
 
 
 	// Traditional WordPress plugin locale filter
-	$locale        = apply_filters( 'plugin_locale',  get_locale(), 'rcp' );
+
+	$get_locale = get_locale();
+
+	if ( $wp_version >= 4.7 ) {
```
pippinsplugins: `global $wp_version` needs to be defined.

Issue #669, opened 2016-08-31
Tables aren't accessible on network activated installations


mindctrl:
If you network activate, go to the first site, and then try to create a level, you'll get an error that it wasn't created. PHP error is:

`WordPress database error Table 'pressmatic.restrict_content_pro' doesn't exist for query SELECT * FROM restrict_content_pro  ORDER BY list_order; made by do_action('restrict_page_rcp-member-levels'), rcp_member_levels_page, rcp_get_subscription_levels, RCP_Levels->get_levels`

If you go to a second site, it works. If you go back to the first site, you see the level you created on the second site, and then it now works on the first site.

We set an empty prefix on network activated installs here: https://github.com/restrictcontentpro/restrict-content-pro/blob/2.6.6/restrict-content-pro.php#L41

Not sure why it works if you go to the second site and back. Haven't looked at it yet. Logging for 2.7 unless you want to fix it up sooner @pippinsplugins.

pippinsplugins:
Let's do it in 2.7 unless we start to see more tickets about it.

pippinsplugins:
This works properly for me now. I think our introduction of `rcp_register_databases()` may have fixed this.

Closed as resolved!

mindctrl:
@pippinsplugins I saw this happen again about 2 days ago on a new customer site. It only happens when network activated, and I think it's due to us setting an empty string for the db prefix in all the `rcp_get_*_name()` functions.

I'm going to open this back up and run another test since I just saw it happen and will close it again if I can't reproduce.

pippinsplugins:
Hmmm, I tested network activation and adding sub levels to multiple sites.

mindctrl:
This happens because when clicking `Network Activate`, the `is_plugin_active_for_network` checks inside `rcp_get_*_db_name` return false. This results in the table names being prefixed with `$wpdb->prefix` during activation, so those are the tables that get created.

But after the plugin is fully activated, `is_plugin_active_for_network` returns true, and `rcp_get_*_db_name` functions return the unprefixed name.

We should probably check for `$network_wide` in `rcp_options_install()` and get the values accordingly.

pippinsplugins:
:+1:

mindctrl:
https://github.com/restrictcontentpro/restrict-content-pro/commit/9b5b83050dd8f340748845d9a4875956c0cb7eaf is one idea for handling this. Works in local tests for both network activated and site-level activated.

[...]

PR #796, opened 2016-11-30
Adjust table creation and upgrade routines when network activated. #669

mindctrl:
#669

### DIFF
--- a/restrict-content-pro.php
+++ b/restrict-content-pro.php
@@ -38,7 +38,7 @@ if ( ! function_exists( 'is_plugin_active_for_network' ) ) {
 function rcp_get_levels_db_name() {
 	global $wpdb;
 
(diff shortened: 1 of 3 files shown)


restrict-content-pro.php
```diff
@@ -56,6 +64,10 @@ function rcp_get_discounts_db_name() {
 
 	$prefix = is_plugin_active_for_network( 'restrict-content-pro/restrict-content-pro.php' ) ? '' : $wpdb->prefix;
```
pippinsplugins: Could we go ahead and change this to `RCP_FILE` to ensure it always works, even if the plugin folder is different?

mindctrl: It looks like we can use `plugin_basename( __FILE__ )`, but unless we create a new constant we'll run it all the time.

mindctrl:
@pippinsplugins I went ahead and pushed a change to use `plugin_basename( RCP_PLUGIN_FILE )`.

Issue #795, opened 2016-11-30
Redirect premium posts when restricted only via taxonomy


mindctrl:
If a post is restricted only via taxonomy, the redirect doesn't happen. It would be nice to support this, as it would make the taxonomy restrictions much more useful and viable with the redirect option. There are some performance concerns with querying the taxonomies and terms connected to a post, but I'd like to dig into it and see if we can come up with a decently performant way to do this.

mindctrl:
@pippinsplugins took a stab at this in `issue/795`. Needs some testing and probably a little bit of error checking. Thoughts on this approach?

pippinsplugins:
See comment on PR regarding using the WP query classes.

mindctrl:
After looking through `WP_Term_Query`, I don't see a way to specify post ID to narrow down the query into one step. I think the approach here is the most performant, but happy to go another route with a WP core method if we can identify one.

pippinsplugins:
Ok this approach seems fine then.

In testing I've found a few things.

1. Terms were coming back as having restrictions even when no restrictions were set. This was due to `None` getting saved in the access level. Fixed in b529dc899ccec3ca76be832684b5088e025d45ff.

2. The `Hide Premium Posts` feature was originally built only to redirect away from _premium_ posts, so content restricted to free members was never affected. In the current implementation, however, this is not the case with taxonomy terms. For example, adding a restriction to a free subscription level will result in the redirect occurring. Basically it's not respecting the `rcp_is_paid_content()` check for terms. This is a bit tricky to fix and, frankly, I'm not sure we should fix it. I actually think we should work on moving towards supporting redirects for _all_ cases where a member does not have access to the content.

3. The current implementation does not properly check if a member has access to terms the content is assigned to before redirecting away from it.

There are two things I think we should do here:

1. Implement a usage of `RCP_Member->can_access( $post_id )` in the redirect function
2. Update `can_access()` to check the taxonomy term restrictions. Right now it doesn't even acknowledge they exist, this is done separately in `rcp_filter_restricted_category_content()`, which is tied to `the_content`. 

Moving the restriction checks on terms to `can_access()` provides several major benefits:

1. Allows us to do redirects for any kind of content, not just paid content.
2. Makes it _significantly_ easier to programmatically check if a member has access to a post.
3. Further reduces confusion in our own code base.

mindctrl:
I agree.

1. We save `None` on post restrictions. It would be great to move away from that too.

[...]

PR #797, opened 2016-11-30
Redirect premium posts when restricted via taxonomy. #795

mindctrl:
WIP: #795


includes/content-filters.php
```diff
@@ -156,6 +156,24 @@ function rcp_is_post_taxonomy_restricted( $post_id, $taxonomy, $user_id = null )
...
+/**
+ * Determines if a post is assigned to any restricted taxonomy terms.
+ *
+ * @since 2.7
+ * @param int $post_id The post ID.
+ * @return boolean True if the post is assigned to any restricted terms, false if not.
+ */
+function rcp_post_has_restricted_taxonomy_terms( $post_id ) {
+	global $wpdb;
+	$term_ids = $wpdb->get_results( $wpdb->prepare( "SELECT term_taxonomy_id FROM {$wpdb->term_relationships} WHERE object_id = %d", absint( $post_id ) ), ARRAY_A );
```
pippinsplugins: Could we use WP_Tax_Query with a meta query to do this instead?

mindctrl: @pippinsplugins I don't see a way to this from `WP_Tax_Query`, but I might be overlooking it. Any ideas?

pippinsplugins: My bad, `WP_Term_Query` https://developer.wordpress.org/reference/classes/wp_term_query/

PR #801, opened 2016-12-04
Add action hook in admin form handler for adding and editing subscriptions

dkjensen:
With the ability to add subscription level meta, there should be a hook during the adding and editing of subscription levels process where the user can add or update level meta. For example, adding a new subscription level field with the `rcp_add_subscription_form` hook.


dkjensen:
It's late, just realized a hook already for this inside the RCP_Level insert() and update() methods.

PR #811, opened 2016-12-08
Update and fix French translation

geoffreybr:
About half of the translation file updated and fixed, until "PayPal API Credentials".


pippinsplugins:
Thanks!

Issue #587, opened 2016-07-12
Add post class to the_content when restricted


michaelbeil:
> Love this plugin, but I was wondering if there's any chance that in your next update you could add a function that adds a class to the body when it's restricted so that it's easier to style the page as needed?

Reference: https://wordpress.org/support/topic/add-body-class-1

nosegraze:
Would it be better to add this to `post_class` rather than `body_class` ? I guess I'm thinking of archive pages.

And maybe two classes:

1. If the content is restricted.
2. If the current user can access the restricted content or not.

mindctrl:
I originally said the same thing about `post_class`. I'm sure people could use either, but I think we should stick with `post_class` since that's what we currently check for restrictions.

We'll want to check what this does to query counts.

nosegraze:
Added a WIP. It doesn't work for taxonomies yet, but I figure we'll wait for when we get that check added to `can_access()` in #795 . Also maybe we should add the taxonomy check to `rcp_is_restricted_content()`

pippinsplugins:
PR looks good so far!

pippinsplugins:
Looks great!

No need to check taxonomy restrictions since we're working on including that in `rcp_user_can_access()` automatically.

PR #813, opened 2016-12-08
WIP: Post classes for restricted content #587

nosegraze:
#587


includes/template-functions.php
```diff
@@ -132,4 +132,35 @@ function rcp_locate_template( $template_names, $load = false, $require_once = tr
...
+ * @param int          $post_id ID of the current post.
+ *
+ * @since 2.7
+ * @return array
+ */
+function rcp_post_classes( $classes, $class = '', $post_id = false ) {
+
+	global $user_ID;
+
+	if ( ! $post_id || is_admin() ) {
```
mindctrl: I don't think we need to test for $post_id here, and we don't need to set it to false in the function param. The `post_class` filter always gets passed `$post->ID`, which would tank before we had a chance to run if the post object wasn't already set up. Is there a scenario you know of where it might not get passed to the filter @nosegraze?

includes/template-functions.php
```diff
@@ -132,4 +132,35 @@ function rcp_locate_template( $template_names, $load = false, $require_once = tr
...
+ * @param array        $classes Array of post classes.
+ * @param string|array $class Additional classes added to the post.
+ * @param int          $post_id ID of the current post.
+ *
+ * @since 2.7
+ * @return array
+ */
+function rcp_post_classes( $classes, $class = '', $post_id = false ) {
+
+	global $user_ID;
```
mindctrl: I'd rather do `$user_id = get_current_user_id();` here instead of the global.... because in 2030 when WordPress moves away from globals and humans have colonized Pluto, we'll be ready!

nosegraze: Hah! Just updated that.

includes/template-functions.php
nosegraze: I'm not aware of any instance that would happen. I think I noticed that EDD did it this way and figured there was a reason lol. But if there's not one I can take that out.

PR #824, opened 2016-12-14
Fixed links broken

fernandotellado:
Links that doesn't work due to wrong quotes imported from poedit fixed


pippinsplugins:
Thanks!

Issue #360, opened 2016-02-24
Failed payment emails


pippinsplugins:
We should add support for failed payments so that customers can be alerted when their payments are failing to be processed.

michaelbeil:
This would be good.

mindctrl:
Just had another request for this.

nosegraze:
Got a request for this today.

mindctrl:
Another request for emailing on failed charges.

pippinsplugins:
I've added in a new `rcp_recurring_payment_failed` hook that is fired by all gateways whenever a recurring payment fails.

Will wait to finish the actual email process until #748 is finished.

pippinsplugins:
@mindctrl could you take over on this now that `RCP_Emails` has been merged?

mindctrl:
👍 will do.

mindctrl:
After letting this run for several more days, I'm seeing an issue we need to address. With the way Stripe (and possibly other gateways) work, invoices keep piling up when expired, at least for some period of time. Looking back at this particular test customer, my failed invoices have been piling up daily since 11/01/16.

In my tests of a daily subscription, each day that passes a new invoice is created and payment attempts are made for the new invoice and several of the past invoices (not all past invoices it seems). This results in a bunch of webhooks coming in for each failed attempt, within seconds of each other, and an email is sent for each one.

In my Stripe account, I have it set up to attempt 3 charges and then cancel.
![image](https://cloud.githubusercontent.com/assets/1254870/21584569/8e7de9fe-d07a-11e6-8558-533fbefc9a0c.png)

However the subscription is "past due", not cancelled, and new invoices keep being created. We should probably deal with this in some way to ensure subscriptions really get cancelled after a period of time, in a separate issue, but I think in the meantime we should put some kind of limit on the number of failed payment emails we send.

I'm reopening this issue to deal with the email part of it.

pippinsplugins:
Was this subscription failing before you set it to 3 charge attempts?

mindctrl:
@pippinsplugins I already had the 3 charge attempts set up before this customer. I've had it that way for as long as I can remember. No idea why Stripe isn't properly cancelling it.

pippinsplugins:
@mindctrl I'm not sure this is an issue we need to address; that's really a Stripe problem.

[...]

PR #833, opened 2016-12-29
Add option to email member when renewal payment fails. #360

mindctrl:
#360

### DIFF
--- a/includes/email-functions.php
+++ b/includes/email-functions.php
@@ -463,3 +463,35 @@ function rcp_email_tag_amount( $member_id = 0, $payment_id = 0 ) {
 function rcp_email_tag_site_name() {
 	return wp_specialchars_decode( get_bloginfo( 'name' ), ENT_QUOTES );
 }
+
+/**
+ * Emails a member when a renewal payment fails.
+ *
+ * @since 2.7
+ * @param object $member  The member (RCP_Member object).
+ * @param object $gateway The gateway used to process the renewal.
+ * @return void
+ */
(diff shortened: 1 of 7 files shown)


includes/email-functions.php
```diff
@@ -462,4 +462,36 @@ function rcp_email_tag_amount( $member_id = 0, $payment_id = 0 ) {
...
+function rcp_email_member_on_renewal_payment_failure( RCP_Member $member, RCP_Payment_Gateway $gateway ) {
+
+	global $rcp_options;
+
+	if ( ! empty( $rcp_options['disable_renewal_payment_failed_email'] ) ) {
+		return;
+	}
+
+	$status    = $member->get_status();
+	$user_info = get_userdata( $member->ID );
```
pippinsplugins: `$member` is an instance of `RCP_Member`, which includes all properties from `WP_User` so this shouldn't be necessary.

mindctrl: @pippinsplugins good catch. copy pasta leftover. updated.

includes/email-functions.php
```diff
@@ -462,4 +462,35 @@ function rcp_email_tag_amount( $member_id = 0, $payment_id = 0 ) {
...
+ */
+function rcp_email_member_on_renewal_payment_failure( RCP_Member $member, RCP_Payment_Gateway $gateway ) {
+
+	global $rcp_options;
+
+	if ( ! empty( $rcp_options['disable_renewal_payment_failed_email'] ) ) {
+		return;
+	}
+
+	$status    = $member->get_status();
```
pippinsplugins: Indention is incorrect here :)

PR #837, opened 2017-01-02
Remove the price select box

ramiy:
Remove the price select box, it is useless, it only resets the price.

rcp subscription levels


mindctrl:
I think I'm fine with this. @pippinsplugins second opinion?

pippinsplugins:
Approved and merged.

PR #852, opened 2017-01-10
Cookie name filter

MikeNGarrett:
Pantheon requires cookies that need to get back to the server to be prefixed. If you add a filter here, it would allow developers the ability to change the cookie name to allow the cookie to bypass caching filters. Docs: https://pantheon.io/docs/cookies/

### DIFF
--- a/includes/login-functions.php
+++ b/includes/login-functions.php
@@ -149,7 +149,7 @@ function rcp_process_lostpassword_reset() {
 	nocache_headers();
 
 	list( $rp_path ) = explode( '?', wp_unslash( $_SERVER['REQUEST_URI'] ) );
-	$rp_cookie = 'rcp-resetpass-' . COOKIEHASH;
+	$rp_cookie = apply_filters( 'rcp_resetpass_cookie_name', 'rcp-resetpass-' . COOKIEHASH );
 
 	// store reset key and login name in cookie & remove from URL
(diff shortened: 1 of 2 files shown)


includes/login-functions.php
```diff
@@ -149,7 +149,7 @@ function rcp_process_lostpassword_reset() {
 	nocache_headers();
 
 	list( $rp_path ) = explode( '?', wp_unslash( $_SERVER['REQUEST_URI'] ) );
-	$rp_cookie = 'rcp-resetpass-' . COOKIEHASH;
+	$rp_cookie = apply_filters( 'rcp_resetpass_cookie_name', 'rcp-resetpass-' . COOKIEHASH );
```
mindctrl: @MikeNGarrett any particular reason for using a different filter name than in #853?

MikeNGarrett: My original thought was to allow two different filters to exist since the cookie names are different. This would allow devs to return different strings, depending on the context, rather than using logic to try and detect the context.

MikeNGarrett:
WordPress is in the middle of a release, so Travis might have some trouble. 

I corrected issue you raised in #853 and closed that issue. I also changed the filter name to be the same since the cookie name is the same.

mindctrl:
@MikeNGarrett thanks! I'm merging this into `release/2.7`.

Issue #830, opened 2016-12-18
Existing users can't subscribe using Stripe Checkout


boluda:
Hi there!

When an existing WordPress user tries to subscribe using the Stripe Checkout he can't. The user has to be deleted, then he can subscribe.

mindctrl:
Followed up in the support ticket you opened. I'll close this for now and reopen when we have more info.

mindctrl:
Confirmed this bug. When an existing user tries to pay for a subscription **while logged out**, Stripe Checkout appears to process the payment, then the registration page reloads with the following errors:
```
Username already taken
Email already registered
```

No charge appears to hit Stripe. We may need to introduce a user/email lookup here. Other gateways show the error before attempting payment and without reloading the page.

pippinsplugins:
Related https://github.com/restrictcontentpro/restrict-content-pro/issues/562

mindctrl:
Let's hold off on testing/merging this just yet. I want to explore another idea.

mindctrl:
@pippinsplugins submitted an alternate idea to `issue-380-alternate`. Doesn't do the inline auto-validation like the first proposal, but uses the existing ajax validation we have and reduces the duplication.

Tested this with Manual, Stripe, PayPal Standard/Pro/Express and it seems ok. Will need to make another pass and also look at Braintree.

Needs another set of eyes too.

mindctrl:
Will have an update here later today for #830, #562, and #855 as they're all connected.

pippinsplugins:
:+1:

pippinsplugins:
All good here!

nosegraze:
Works perfectly with `[register_form]`. I'm not sure if it was meant to be addressed, but this is still an issue with `[register_form_stripe]` though. The modal makes it look like it's been paid but it hasn't been.

mindctrl:
Merged.

Note: we found no event to tie into when someone enters an email in the modal when using just the shortcode version, so we can't do any JS validation before the form submits at this time.

PR #841, opened 2017-01-05
Add registration form ajax validation. #830 #562

mindctrl:
#830, #562

Issue #562, opened 2016-06-28
Stripe Checkout: Submit Payment should be disabled until form validations pass


carter2422:
The **Submit Payment** button with Stripe Checkout should really be disabled until all forms are filled out. Currently it can lead to a very confusing user flow if the user submits payment without filling out their details as it _appears_ as a payment has been submitted (it hasn't). The main reason for this is the green checkmark animation on the checkout form suggests success.

mindctrl:
Punting this to 2.7 so we can spend time verifying everything, like matching passwords, proper email, username exists, etc.

pippinsplugins:
PR works great!

@nosegraze could you give it a test as well please?

nosegraze:
Works perfectly after the update!

PR #857, opened 2017-01-12
Validate form data in Stripe Checkout before opening modal. #830, #562, #855

### LINKED ISSUE #855: Invalid integer error when using Alipay
We need to pass the `data-amount` to the Checkout window for Alipay to work. From Stripe support:
> If you're using Alipay with Checkout you absolutely must include a 'data-amount' parameter as part of your Checkout window. This is because Alipay needs to know beforehand how much to allow the user to be charged for. This isn't required for card payments and thus why a card payment works but not Alipay.

mindctrl:
This is an alternative to #841, which I think we shouldn't use now. Though I'd like to revisit the data validation on the fly in the near future.

This PR slightly reworks some of the internals of the gateways. This was largely due to the registration form being forcefully submitted in register.js. With this:
- the gateway scripts control when the form gets submitted. This was really necessary for Stripe Checkout, and lays the groundwork for future improvements. This happens on the `rcp_register_form_submission` trigger, which only fires when form data validation has passed. This fixes issues #830 and #562.

- register.js only submits the form if the total is $0, since the gateway scripts don't run in that case.

- `data-amount` is now passed to the Stripe Checkout modal, which is a requirement for Alipay to work. This fixes #855.

- bonus, the registration form is unblocked if someone manually closes the Stripe Checkout modal.

@pippinsplugins This needs another pass to clean up a couple things, but wanted to get an opinion on this approach before moving much further.

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-paypal.php
+++ b/includes/gateways/class-rcp-payment-gateway-paypal.php
@@ -498,7 +498,12 @@ class RCP_Payment_Gateway_PayPal extends RCP_Payment_Gateway {
 	public function scripts() {
 		wp_add_inline_script(
 			'rcp-register',
-			"jQuery('body').on('rcp_register_form_submission', function(e, response, form, submission_form) { submission_form.submit(); });"
+			"jQuery('body').on('rcp_register_form_submission', function(e, response, form, submission_form) {
+				if (form.data.gateway !== 'paypal') {
+					return;
+				}
+				submission_form.submit();
+			});"
 		);
 	}
 
--- a/includes/gateways/class-rcp-payment-gateway-paypal.php
+++ b/includes/gateways/class-rcp-payment-gateway-paypal.php
@@ -495,4 +495,11 @@ class RCP_Payment_Gateway_PayPal extends RCP_Payment_Gateway {
 
 	}
 
+	public function scripts() {
+		wp_add_inline_script(
+			'rcp-register',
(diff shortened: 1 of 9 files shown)


includes/gateways/class-rcp-payment-gateway-paypal.php
```diff
@@ -495,4 +495,16 @@ public function process_webhooks() {
 
 	}
 
+	public function scripts() {
+		wp_add_inline_script(
+			'rcp-register',
+			"jQuery('body').on('rcp_register_form_submission', function(e, response, form, submission_form) {
+				if (form.data.gateway !== 'paypal') {
```
pippinsplugins: These could all be done in a base method. It just needs to inject the gateway ID into the JS.

pippinsplugins:
Fairly certain this should work. We'll need to be sure to test it with at least one or two of the third party gateways.

mindctrl:
@pippinsplugins adjusted this to default to register.js submitting it, and passing a new attribute to gateway supports. Reduces duplication and should ensure third party gateways still work the same.

includes/gateways/class-rcp-payment-gateway-stripe-checkout.php
```diff
@@ -11,6 +11,11 @@
 
 class RCP_Payment_Gateway_Stripe_Checkout extends RCP_Payment_Gateway_Stripe {
 
+	public function init() {
```
pippinsplugins: Let's PHPdoc this please

PR #870, opened 2017-01-17
a11y - Admin Headings

ramiy:
From an a11y point of view, you should always have `<h1>` heading in all the backend pages.

Related: https://core.trac.wordpress.org/ticket/31650

This PR replaces `H2` headings with `H1`, and other heading accordingly.


mindctrl:
This had a merge conflict after basing it against `release/2.7`. Resolved that and committed here: https://github.com/restrictcontentpro/restrict-content-pro/commit/8b1c7f9193bef64d8125b35e51ad16380669e9e9. Thanks @ramiy.

PR #875, opened 2017-01-17
Introduce rcp_get_currency_symbol(). Use it in rcp_currency_filter().

leewillis77:
This PR introduces rcp_get_currency_symbol(), and updates rcp_currency_filter() to use it. This makes it easier for other plugins to fetch the configured "symbol" if they need it. 

rcp_currency_filter() is more of a mess than I'd like, but mainly to preserve the existing filter calls.


includes/misc-functions.php
```diff
@@ -72,73 +72,20 @@ function rcp_currency_filter( $price ) {
...
-	endif;
+	$symbol   = rcp_get_currency_symbol();
+	if ( 'before' === $position ) {
+		if ( $symbol === $currency ) {
+			return apply_filters( 'rcp_' . strtolower( $currency ) . '_currency_filter_before', $symbol . ' ' . $price, $currency, $price );
+		} else {
+			return $symbol . $price;
+		}
+	} else {
+		if ( $symbol === $currency ) {
```
mindctrl: @leewillis77 what's the reason for wrapping these two returns in a `$symbol === $currency` check? If I'm reading it right, that means this filter will rarely, if ever, run because they are generally not equal.

leewillis77: If you take a look at the original code, specifically:

https://github.com/restrictcontentpro/restrict-content-pro/pull/875/files#diff-396f42fbf79a5fb9b778b84d7084c54bL138
and then
https://github.com/restrictcontentpro/restrict-content-pro/pull/875/files#diff-396f42fbf79a5fb9b778b84d7084c54bL140

then you should see that in the case where we resolve a currency string, e.g. USD into a symbol, the code returns the symbol with price immediately without passing it through a filter.

The filter is only called in the case where the string couldn't be resolved into a symbol (The default case on the switch falls through, and doesn't return immediately).

So - this logic preserves the behaviour of the original code with respect to when filters are / aren't called.

Does that make sense?

leewillis77:
Hi; 

What's the status of this. Is it OK in principal (in which case I can update the PR since it looks like there are conflicts now), or is it unlikely to go anywhere?

mindctrl:
@leewillis77 sorry, I lost track of this since it's not connected to an issue. I'll make an issue for it and see if we can get it in 2.9.1 or so.

leewillis77:
No problem - thanks!

Do you want me to fix up this PR? or would you prefer to take a look yourself?

mindctrl:
https://github.com/restrictcontentpro/restrict-content-pro/pull/1341

PR #878, opened 2017-01-17
Fix programmatic discount code creation errors in Stripe gateway.

leewillis77:
The Stripe gateway code attempts to create discount codes in Stripe when
one is created in RCP. However, it does not use the information made
available to the hook, instead grabbing data out of $_POST. This fails
when the discount code is being created programatically.

This change passes in the discount args, and uses them instead of $_POST.


mindctrl:
Thanks @leewillis77!

PR #891, opened 2017-01-24
Update process-data.php

davexpression:
Added an action hook for the function that adds a new subscription


mindctrl:
@davexpression thanks for the PR. Can you talk about your use case here?

includes/process-data.php
```diff
@@ -139,6 +139,8 @@ function rcp_process_data() {
 				} else {
 					delete_user_meta( $user->ID, 'rcp_recurring' );
 				}
+				
+				do_action( 'rcp_add_subscription', $user->ID, $member, $subscription );
```
mindctrl: @davexpression we'll want to add more context to this hook name I think... maybe `rcp_add_subscription_to_user`, for example.

davexpression:
Hello John.

I had some existing members I manually added via the Restrict Members page.
I noticed their status in the usermeta is "free". However, when I created a
new user via the frontend, their status is active and "is_trialling" is set
to "yes". Why am I bothered about the status. Simple.

If you run rcp_is_active() or rcp_is_trialing(), it will return false for
members with a "free" status. I had a lot of code I wanted to apply to free
trial members whether added via frontend or backend. So, the hook enabled
me to do so like this:

function admin_user_subscription($user, $member, $subscription){
  if ($subscription->price == 0){
  $member->set_status ('active');
  update_user_meta( $user, 'rcp_is_trialing', 'yes' );
update_user_meta( $user, 'rcp_has_trialed', 'yes' );
rcp_email_subscription_status( $user, 'trial' );
      update_user_meta( $user, 'custome_meta', '123456' );
  }
}

add_action( 'rcp_add_subscription',  'admin_user_subscription', 10, 3 );

davexpression:
Yea, great

mindctrl:
@davexpression thanks for the explanation.

RCP should only set a member's status to "free" if you assign them a level that has a price of 0 and a duration of 0. That's intended behavior, as the level they're assigned to is essentially free forever, hence the "free" status.

If you assigned members to a level that costs 0 but has a duration that is not 0, they should get set as "active". If they are not being set as active, that's a bug and should be logged and fixed separately.

PR #883, opened 2017-01-19
WIP: Fix missing/broken PHPDocs #879

nosegraze:
#879

PR #911, opened 2017-02-02
Adding email to admin cancellation notification

donnfelker:
This improves the email by allowing admins to have the email to directly reply to folks after they cancel. Otherwise, you have to log into the admin and find the users email before you can email them directly. This commit adds the email to eliminate that extra step. 

I'm aware that you can have RCP send an email upon cancellation and I've been doing so, but I've found that sometimes I need to follow up with these folks for other reasons. Having direct access to their email via the cancellation email would help this workflow a lot.


mindctrl:
I changed this PR to `release/2.7` instead of `master`, however there's a merge conflict now. I'll address that after https://github.com/restrictcontentpro/restrict-content-pro/pull/883 is merged.

Issue #915, opened 2017-02-03
URLs are not clickable in email bodies with RCP_Emails


mindctrl:
A URL entered into the body of an email is not clickable. This will be a problem for things like Group Accounts invite links, and maybe password reset links (haven't tested that one). We may need to run it through `make_clickable`. 

![image](https://cloud.githubusercontent.com/assets/1254870/22597928/a309816e-e9ff-11e6-8488-eb9459ea3bb6.png)

PR #916, opened 2017-02-03
Run HTML message body through make_clickable(). #915

mindctrl:
#915

### DIFF
--- a/includes/class-rcp-emails.php
+++ b/includes/class-rcp-emails.php
@@ -240,6 +240,8 @@ class RCP_Emails {
 			return apply_filters( 'rcp_email_message', wp_strip_all_tags( $message ), $this );
 		}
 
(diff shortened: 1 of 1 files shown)


nosegraze:
It doesn't really matter of course, but I wonder if it would be better to stick it in the `text_to_html()` method. Like:

```
$message = make_clickable( wpautop( $message ) );
```

It's still relevant to what's happening in that method IMO (text to HTML) and means the check only has to happen once.

mindctrl:
@nosegraze I tried that and it didn't work for the Group Account emails. I didn't dig deeper to figure out why.

mindctrl:
Hmm it might have been on an older Group Accounts version. Not sure. I'll look at it again later.

nosegraze:
Works perfectly for me!

Issue #849, opened 2017-01-09
Set display name to username if first & last name aren't filled out


nosegraze:
When creating user accounts we set the display name as follows:

`'display_name'    => $user_data['first_name'] . ' ' . $user_data['last_name'],`

But neither first name nor last name are actually required in the form, so if neither one is filled out then the display name is just totally blank, which is a bit weird. I'm thinking we should make that fall back to the username if both fields are empty.

Had someone with this problem here: https://secure.helpscout.net/conversation/299269629/50084/?folderId=1100052

pippinsplugins:
:+1:

nosegraze:
PR: #938

PR #938, opened 2017-02-09
Set display name to username if first and last name fields are empty. #849

nosegraze:
#849

### DIFF
--- a/includes/registration-functions.php
+++ b/includes/registration-functions.php
@@ -140,7 +140,7 @@ function rcp_process_registration() {
 				'user_email'      => $user_data['email'],
 				'first_name'      => $user_data['first_name'],
 				'last_name'       => $user_data['last_name'],
-				'display_name'    => $user_data['first_name'] . ' ' . $user_data['last_name'],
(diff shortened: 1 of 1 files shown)


includes/registration-functions.php
```diff
@@ -140,7 +140,7 @@ function rcp_process_registration() {
 				'user_email'      => $user_data['email'],
 				'first_name'      => $user_data['first_name'],
 				'last_name'       => $user_data['last_name'],
-				'display_name'    => $user_data['first_name'] . ' ' . $user_data['last_name'],
+				'display_name'    => ! empty( trim( $user_data['first_name'] . ' ' . $user_data['last_name'] ) ) ? $user_data['first_name'] . ' ' . $user_data['last_name'] : $user_data['login'],
```
mindctrl: @nosegraze using `trim()`, or any function, inside `empty()` fails on PHP versions less than 5.5. The tests failed for those versions because of this. So so meh. I'm ready to stop supporting old PHP personally, but can you update this?

nosegraze: Derp. So just put it in a variable first?

mindctrl: Yeah, you can trim it after the `?` too to avoid creating a new var I think.

Issue #899, opened 2017-01-30
Better filters for members and payments export data


nosegraze:
Right now our members export code looks like this:

```php
$members      = rcp_get_members( $status, $subscription, $offset, $number );

if( $members ) :
	foreach ( $members as $member ) {

		$member = new RCP_Member( $member->ID );

		$discounts = get_user_meta( $member->ID, 'rcp_user_discounts', true );
		if( ! empty( $discounts ) && is_array( $discounts ) && ! $discounts instanceof stdClass ) {
			foreach( $discounts as $key => $code ) {
				if( ! is_string( $code ) ) {
					unset( $discounts[ $key ] );
				}
			}
			$discounts = implode( ' ', $discounts );
		}

		$data[] = array(
			'user_id'          => $member->ID,
			'user_login'       => $member->user_login,
			'user_email'       => $member->user_email,
			'first_name'       => $member->first_name,
			'last_name'        => $member->last_name,
			'subscription'     => $member->get_subscription_id(),
			'subscription_key' => $member->get_subscription_key(),
			'expiration'       => $member->get_expiration_date(),
			'status'           => $member->get_status(),
			'discount_codes'   => $discounts,
			'profile_id'       => $member->get_payment_profile_id(),
			'is_recurring'     => $member->is_recurring()
		);

	}
endif;

$data = apply_filters( 'rcp_export_get_data', $data );
$data = apply_filters( 'rcp_export_get_data_' . $this->export_type, $data );
```

The only filters we have are _after_ the `foreach` loop. The same for payments. That makes it kind of awkward to work with if you actually want to include extra member/payment data in the export. You'd have to loop through all the data all over again.

We should filter the individual user/payment data inside the foreach loop so more data can be added to each row.

nosegraze:
PR: #939

PR #939, opened 2017-02-09
Add filters for export line items. #899

nosegraze:
#899


mindctrl:
@nosegraze looks good. Thoughts on changing `line` to `row` in the filter name, to match the column/row naming convention?

pippinsplugins:
I agree with changing `line` to `row`.

nosegraze:
Updated. That's much better! :D

Issue #860, opened 2017-01-13
Issues with rcp_get_current_url() when using reverse proxies


nosegraze:
`rcp_get_current_url()` doesn't work well when using a reverse proxy behind another reverse proxy (i.e. PHP behind NGINX behind Varnish). The usage of `$_SERVER["SERVER_PORT"]` has caused some quirks for me when doing things like cancelling subscriptions.

I want to play around with this a bit and see if I can come up with a better way of doing it.

nosegraze:
PR: #940

Any issues with this? For `rcp_get_current_url()` I copied chunks from EDD's version and it seems to work well for me. The main thing that was causing issues was processing cancellations since that runs on `init`, when none of the conditional statements work yet. Changing that to `template_redirect` solves it.

mindctrl:
This works well. It fixed an issue for me while testing https://github.com/restrictcontentpro/restrict-content-pro/pull/937.

PR #940, opened 2017-02-09
Rework rcp_get_current_url() and change action for processing cancellations #860

nosegraze:
#860

### DIFF
--- a/includes/misc-functions.php
+++ b/includes/misc-functions.php
@@ -281,16 +281,21 @@ function rcp_get_current_url() {
 
 	else :
 
-		$current_url = 'http';
-		if ( is_ssl() ) $current_url .= "s";
+		global $wp;
 
-		$current_url .= "://";
+		if( get_option( 'permalink_structure' ) ) {
 
-		if ( $_SERVER["SERVER_PORT"] != "80" ) {
-			$current_url .= $_SERVER["SERVER_NAME"] . ":" . $_SERVER["SERVER_PORT"] . $_SERVER["REQUEST_URI"];
+			$base = trailingslashit( home_url( $wp->request ) );
+
 		} else {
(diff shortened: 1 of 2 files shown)


includes/misc-functions.php
```diff
@@ -281,17 +281,22 @@ function rcp_get_current_url() {
...
+		if( get_option( 'permalink_structure' ) ) {
+
+			$base = trailingslashit( home_url( $wp->request ) );
 
-		if ( $_SERVER["SERVER_PORT"] != "80" ) {
-			$current_url .= $_SERVER["SERVER_NAME"] . ":" . $_SERVER["SERVER_PORT"] . $_SERVER["REQUEST_URI"];
 		} else {
-			$current_url .= $_SERVER["SERVER_NAME"] . $_SERVER["REQUEST_URI"];
+
+			$base = add_query_arg( $wp->query_string, '', trailingslashit( home_url( $wp->request ) ) );
```
mindctrl: I think this needs to be escaped since it's not passing the second param. https://make.wordpress.org/plugins/2015/04/20/fixing-add_query_arg-and-remove_query_arg-usage/#comment-42117

nosegraze: We do send the URL here though in the third param: `trailingslashit( home_url( $wp->request ) )` I'm guessing when he said "second param" he was referring to the second param when using an array for the first one. But we're using two params (key/value) rather than an array.

We pretty much always escape later whenever we use the actual function (i.e. `echo esc_url( rcp_get_current_url() )`). We could update the ones where we don't. I'm just wondering if that's better since there are some cases where we'd want to use `esc_url()` and others when we might want `esc_url_raw()` (like the redirect in cancellation).

mindctrl: Yeah I think we should just ensure we're escaping where we use it.

PR #965, opened 2017-02-14
Improvements to admin email notifications #7

nosegraze:
#7

### DIFF
--- a/includes/email-functions.php
+++ b/includes/email-functions.php
@@ -28,8 +28,8 @@ function rcp_email_subscription_status( $user_id, $status = 'active' ) {
 	$site_name     = stripslashes_deep( html_entity_decode( get_bloginfo('name'), ENT_COMPAT, 'UTF-8' ) );
 
 	$admin_emails  = ! empty( $rcp_options['admin_notice_emails'] ) ? $rcp_options['admin_notice_emails'] : get_option('admin_email');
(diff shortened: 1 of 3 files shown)


includes/email-functions.php
```diff
@@ -27,9 +27,9 @@ function rcp_email_subscription_status( $user_id, $status = 'active' ) {
 	$admin_message = '';
 	$site_name     = stripslashes_deep( html_entity_decode( get_bloginfo('name'), ENT_COMPAT, 'UTF-8' ) );
 
-	$admin_emails   = array();
-	$admin_emails[] = get_option('admin_email');
-	$admin_emails   = apply_filters( 'rcp_admin_notice_emails', $admin_emails );
+	$admin_emails  = ! empty( $rcp_options['admin_notice_emails'] ) ? $rcp_options['admin_notice_emails'] : get_option('admin_email');
+	$admin_emails  = array_map( 'sanitize_email', explode( ',', $admin_emails ) );
+	$admin_emails  = apply_filters( 'rcp_admin_notice_emails', $admin_emails );
```
mindctrl: I think we should run the `sanitize_email` bit after the filter, to make sure bad data doesn't get in after it's sanitized.

nosegraze: I'll update that.

Issue #831, opened 2016-12-22
Email admin when manual payment received


mindctrl:
Right now admins don't get notified of new manual payments. After a brief chat in Slack we decided to trigger an email from inside the manual payments gateway.

nosegraze:
PR: #975

PR #975, opened 2017-02-15
Email admin when manual payment is received. #831

nosegraze:
#831


includes/gateways/class-rcp-payment-gateway-manual.php
```diff
@@ -59,7 +59,27 @@ public function process_signup() {
 		);
 
 		$rcp_payments = new RCP_Payments();
-		$rcp_payments->insert( $payment_data );
+		$payment_id   = $rcp_payments->insert( $payment_data );
```
pippinsplugins: Could we add a hook here and then move this code to the `email-functions.php` to keep it all in one place?

nosegraze: Will do!

PR #985, opened 2017-02-17
Pass $event to rcp_stripe_charge_succeeded hook

tnorthcutt:
I don't _think_ there's any downside to this, it should be fully backwards compatible, and it allows for using the stripe event data to inform what actions an extension might want to take. Probably a bit of an edge case in terms of usefulness, but it'd be handy for something I'm working on right now.

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-stripe.php
+++ b/includes/gateways/class-rcp-payment-gateway-stripe.php
@@ -544,8 +544,8 @@ class RCP_Payment_Gateway_Stripe extends RCP_Payment_Gateway {
 					// failed payment
 					if ( $event->type == 'charge.failed' ) {
 
-						do_action( 'rcp_recurring_payment_failed', $member, $this );
-						do_action( 'rcp_stripe_charge_failed', $invoice );
+						do_action( 'rcp_recurring_payment_failed', $member, $this, $event );
+						do_action( 'rcp_stripe_charge_failed', $payment_event, $event );
 
 						die( 'rcp_stripe_charge_failed action fired successfully' );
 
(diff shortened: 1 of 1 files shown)


mindctrl:
This seems ok to me. We could even pass it to the other actions in the same method, I think.

pippinsplugins:
I'm good with this.

tnorthcutt:
Sounds good, I'll edit to include other actions as well.

includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -564,7 +564,7 @@ public function process_webhooks() {
 
 					}
 
-					do_action( 'rcp_stripe_' . $event->type, $payment_event );
+					do_action( 'rcp_stripe_' . $event->type, $payment_event, $event );
```
pippinsplugins: `$payment_event` is the same thing as `$event` here I believe.

includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -544,8 +544,8 @@ public function process_webhooks() {
 					// failed payment
 					if ( $event->type == 'charge.failed' ) {
 
-						do_action( 'rcp_recurring_payment_failed', $member, $this );
-						do_action( 'rcp_stripe_charge_failed', $payment_event );
+						do_action( 'rcp_recurring_payment_failed', $member, $this, $event );
+						do_action( 'rcp_stripe_charge_failed', $payment_event, $event );
```
pippinsplugins: `$payment_event` is the same thing as `$event` here I believe.

includes/gateways/class-rcp-payment-gateway-stripe.php
tnorthcutt: In these cases `$payment_event` is `$event->data->object`. Obviously this becomes a bit redundant and it'd be nice to just pass the entire `$event`, but since that breaks backcompat I opted to go with the redundancy.

tnorthcutt:
@pippinsplugins you good with merging this as-is?

pippinsplugins:
Yep, merging!

PR #992, opened 2017-02-17
Add filter for discount codes

andreamaiolo:
Hi,
we need to filter the array returned by the get_discounts in RCP_Discount class.
Please merge this PR.
Thanks


mindctrl:
@andreamaiolo what's your use case for this?

mindctrl:
We're not consistent in this regard, but I'd prefer any new filters go into helper functions and not the database class. When I use a database class, I want a real view of the data in the db, not a modified version of it.

andreamaiolo:
Hi @mindctrl,
I'm working on a plugin and I need to filter the discount codes. My plugin generates discount codes and I prefer to show the generated codes in a separate table in the admin area of WordPress.

PR #993, opened 2017-02-19
Remove BuddyPress admin bar.

MikeNGarrett:
BuddyPress maintains a bit of control over when they show the admin bar. If the a user decides to remove the admin bar, it would be a good idea to apply this to BuddyPress as well.


mindctrl:
We have no formal integrations with BuddyPress, and this isn't structured in the format we use for integrations. I vote this goes into the BuddyPress add-on by @tannerm instead.

Issue #848, opened 2017-01-09
Option to apply discounts to signup fees


nosegraze:
Got a requested for this. Currently discounts are applied to the subscription price but not signup fees. This was suggested as an extra settings option.

MikeNGarrett:
I would love to see this because applying discounts should apply for both subscription fee and signup fee. It's really difficult to understand without being able to discover this through code.

MikeNGarrett:
PR #996 is an attempt to look at applying discount to both signup fee and subscription fee. 

## Here's what I found: 
#483 is addressing an initial discount vs subscription discount. 
This will change discount processing a bit since it affects how discounts are calculated and applied to initial charges and subscription charges.

Since each payment gateway handles signup fee vs subscription fee in a different way, it seems like it would be best to calculate this before it gets to the gateway. 

In addition, the current functionality does not apply a discount to the signup fee, so we would need a way to make this the default.

It also appears that using a single discount amount for both the subscription amount and signup fee would further complicate the code and may lead to undesirable conditions.

## So,
To accomplish this, we need a separate section in the discount settings that specifically applies to a discount for the signup fee. We could replicate the current discount amount vs percentage interface. Then, we would have to add this additional calculation to the signup fee and handle situations where both the signup fee and subscription fee are 0.

mindctrl:
I think for our own sanity any attempts to do this should come after https://github.com/restrictcontentpro/restrict-content-pro/issues/668 is resolved.

rumspeed:
+1 on this option

PR #996, opened 2017-02-20
Apply discount to signup fee. #848

MikeNGarrett:
#848 reads,
> Got a requested for this. Currently discounts are applied to the subscription price but not signup fees. This was suggested as an extra settings option.

I’m sure I’m oversimplifying this, but I thought I would submit a PR to get the ball rolling. The premise here is to use anything leftover from the discount to apply to the signup fee.

If this is the right direction and generally accepted for further development and testing, I anticipate adding a field to discounts to use the old vs new functionality.

### DIFF
--- a/includes/class-rcp-registration.php
+++ b/includes/class-rcp-registration.php
@@ -302,9 +302,9 @@ class RCP_Registration {
 		}
 
 		// make sure the discount is not > 100%
-		if ( 0 > $total ) {
-			$total = 0;
-		}
+		// if ( 0 > $total ) {
+		// 	$total = 0;
+		// }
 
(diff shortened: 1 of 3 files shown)


MikeNGarrett:
After some testing, it appears the signup fee is still being passed to the gateway. I'm working on that next.

MikeNGarrett:
Would it make more sense to have a separate field set in the discounts settings for the signup fee? I spent this morning going through the gateway setup and it seems like gateways handle signup fee vs subscription fee in different ways. This could simplify the process of getting the total before sending this information over to the gateway.

pippinsplugins:
@MikeNGarrett could we move the discussion over to the issue #848? It's easier for us to keep approach discussions on the issue itself so everything is in one place. Thanks!

pippinsplugins:
PR can stay open, just need to have the discussion on the issue :)

Issue #763, opened 2016-11-11
Option to show excerpts by default


nosegraze:
I've gotten two requests now to have a way to enable excerpts for non-members by default without having to select the option each time.

opt1k4l:
would love to see this too.  looked all over for it in the documentation and then stumbled upon this enhancement.

nosegraze:
I got another request this week and I've been thinking about how to approach this. Maybe we should add a new global "Content Excerpts" setting that has a dropdown with three options:

1. Always show excerpts
2. Never show excerpts
3. Decide for each post individually

If 1 or 2 is chosen then we hide the checkbox from the post metabox (and maybe show text instead reminding them of their chosen setting). If they choose 3 then we show the checkbox as we do now.

pippinsplugins:
I like that approach.

nosegraze:
PR: #997

mindctrl:
PR looks good. A thought about the UI on the post edit screen.

When the global option is set to show/hide, you get a sentence like this:

![image](https://cloud.githubusercontent.com/assets/1254870/23131355/b01022fe-f757-11e6-8334-63f32d46a77b.png)

It looks like a description instead of an option, which is ok. But what do you think about something like this?

![image](https://cloud.githubusercontent.com/assets/1254870/23131394/d1eb15fa-f757-11e6-884f-e057b74cc415.png)

In that example, the global setting is enabled. The checkbox would be checked, but disabled, to keep the UI layout consistent and to show at a quick glance what the setting is. If it were disabled, it would be unchecked and disabled. The tooltip could reveal the hint about the global setting.

Thoughts?

nosegraze:
Yes, I love that! That's a much better idea.

mindctrl:
I think you'd need to check the global setting during save, the avoid saving the individual meta value if globally disabled/enabled... just to avoid adding more post meta when it's not needed.

pippinsplugins:
Love it.

mindctrl:
Works great. Made some minor adjustments and merged.

PR #997, opened 2017-02-20
Add global setting for content excerpts. #763

nosegraze:
#763

### DIFF
--- a/includes/admin/metabox-view.php
+++ b/includes/admin/metabox-view.php
@@ -17,7 +17,7 @@ $set_level         = is_array( $sub_levels ) ? '' : $sub_levels;
 $access_level      = get_post_meta( get_the_ID(), 'rcp_access_level', true );
 $access_level      = is_numeric( $access_level ) ? absint( $access_level ) : '';
 $content_excerpts  = isset( $rcp_options['content_excerpts'] ) ? $rcp_options['content_excerpts'] : 'individual';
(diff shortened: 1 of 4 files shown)


includes/admin/metabox-view.php
```diff
@@ -86,10 +87,24 @@
...
-				<input type="checkbox" name="rcp_show_excerpt" id="rcp-show-excerpt" value="1"<?php checked( true, $show_excerpt ); ?>/>
-				<?php _e( 'Show excerpt to members without access to this content.', 'rcp' ); ?>
-			</label>
+			<?php if ( 'always' == $rcp_options['content_excerpts'] ) :
+				printf(
+					__( 'An excerpt will be shown to members without access to this content. You can change this behavior in %sRestrict > Settings > Misc%s.', 'rcp' ),
+					'<a href="' . esc_url( admin_url( 'admin.php?page=rcp-settings#misc' ) ) . '">',
+					'</a>'
+				);
+			elseif ( 'never' == $rcp_options['content_excerpts'] ) :
```
mindctrl: When checking out this branch, I get debug notices on the new post screen for lines 90 and 96. This is because I haven't saved the global settings. We'll need to account for that for people who upgrade and may not touch the settings for a while.

nosegraze: Updated.

Issue #771, opened 2016-11-16
Add option to remove all data when uninstalling plugin


mindctrl:
When uninstalling RCP, we should have an option that allows for removing all plugin data, tables, etc. It should default to off, so that people don't accidentally remove all data. EDD has an option like this.

nosegraze:
PR: #1009

pippinsplugins:
@nosegraze Could you update the `release/2.8` branch with `master` show the PR is clean and limited to just the relevant changes?

nosegraze:
I'm off my game today.

Should be fixed now.

mindctrl:
I updated this PR to use get_option directly instead of calling the `$rcp_options` global. The global wasn't accessible on my site during uninstall for some reason.

I also updated the `DROP TABLE` stuff to use the table name helper functions. Since those helper functions have filters and the names aren't predictable, I ran them through `esc_sql` to at least ensure they don't have funky data - it appears you can't pass table names through `prepare()`, as it quotes the names and results in invalid SQL. This is the first time I've needed to use this approach, so if anyone knows of a better way I'd love to know about it.

nosegraze:
Makes sense. It's the same for why you can't use `prepare()` for order/orderby values.

I'm good with it!

PR #1009, opened 2017-02-21
Add option to remove all data on uninstall. #771

nosegraze:
#771

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -25,7 +25,7 @@ include_once( 'restrict-content-pro.php' );
 
 global $wpdb, $rcp_options;
 
-if( isset( $rcp_options['uninstall_on_delete'] ) ) {
+if( isset( $rcp_options['remove_data_on_uninstall'] ) ) {
 
 	// Delete all post meta.
 	$wpdb->query( "DELETE FROM $wpdb->postmeta WHERE meta_key LIKE 'rcp\_%'" );
--- a/uninstall.php
+++ b/uninstall.php
@@ -51,10 +51,18 @@ if( isset( $rcp_options['remove_data_on_uninstall'] ) ) {
 	$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE 'rcp\_%'" );
 
 	// Remove all database tables.
-	$wpdb->query( "DROP TABLE IF EXISTS " . $wpdb->prefix . "rcp_discounts" );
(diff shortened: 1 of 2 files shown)


uninstall.php
```diff
@@ -0,0 +1,63 @@
...
+
+	// Clear scheduled cron events.
+	wp_clear_scheduled_hook( 'rcp_expired_users_check' );
+	wp_clear_scheduled_hook( 'rcp_send_expiring_soon_notice' );
+	wp_clear_scheduled_hook( 'rcp_check_member_counts' );
+
+	// Remove all plugin settings.
+	$options = array( 'rcp_settings', 'rcp_db_version', 'rcp_discounts_db_version', 'rcp_payments_db_version', 'rcp_is_installed', 'rcp_version', 'rcp_license_status' );
+	foreach( $options as $option) {
+		delete_option( $option );
```
pippinsplugins: Let's do this as a single query directly to $wpdb for any option `LIKE 'rcp\_%'`

nosegraze: Updated!

uninstall.php
```diff
@@ -0,0 +1,68 @@
...
+	$wpdb->query( "DELETE FROM $wpdb->options WHERE option_name LIKE 'rcp\_%'" );
+
+	// Remove all database tables.
+	$wpdb->query( $wpdb->prepare( "DROP TABLE IF EXISTS %s", rcp_get_discounts_db_name() ) );
+	$wpdb->query( $wpdb->prepare( "DROP TABLE IF EXISTS %s", rcp_get_payments_db_name() ) );
+	$wpdb->query( $wpdb->prepare( "DROP TABLE IF EXISTS %s", rcp_get_payment_meta_db_name() ) );
+	$wpdb->query( $wpdb->prepare( "DROP TABLE IF EXISTS %s", rcp_get_level_meta_db_name() ) );
+	$wpdb->query( $wpdb->prepare( "DROP TABLE IF EXISTS %s", rcp_get_levels_db_name() ) );
+
+	// Remove all options.
```
nosegraze: Options are already removed further up.

mindctrl: Ah, didn't see that. But it's not working. I need to run some more tests.

mindctrl: The global $rcp_options isn't available at this time, at least for me, which is why nothing was getting deleted in my tests. Manually getting it with get_option works for me though. Going to tinker with this a little more later tonight.

PR #1013, opened 2017-02-21
Add is_cancelled shortcode

meadows19:
Useful for targeting users who have cancelled but whose accounts have not yet expired.


nosegraze:
Haven't tested it yet, but just to be clear I'm pretty sure this would include both cancelled users who are expired, and cancelled users who are not expired (since we don't update their status to expired if they deliberately cancel).

meadows19:
I'm sorry, I'm not sure I follow.

I had thought it works like this:

1) A user signs up to a one month recurring membership on Feb 1st.
2) They then cancel using the cancel link on Feb 2nd -- their status is changed to 'cancelled'.
3) They continue to have access to restricted content until Feb 28th, when their status is changed to 'expired'. They lose access to restricted content.

The aim here is for this shortcode to target the user between stages 2 and 3.

What are the other scenarios where the user status is marked as 'cancelled' but not as 'expired'?

nosegraze:
We never update a cancelled member's status to "expired". The expired status is reserved for people who haven't cancelled but have let their membership expire through other means (failure to manually renew when not recurring, or failed recurring payment).

In our checks to see if a user has access, we look at the actual expiration date value for users marked as cancelled. But even when it has expired, we never update their status from cancelled to expired. (They don't have access to anything due to the expiration date value, but their status doesn't change.)

If you specifically want to target cancelled members who have not expired, you'd need something like this:

```php
if( 'cancelled' === rcp_get_status( $user_ID ) && ! rcp_is_expired( $user_ID ) {
    return do_shortcode( $content );
}
```

Though I wonder if it might make sense to have an attribute in the shortcode so people can decide if they want that extra check in there or not.

meadows19:
Thanks for the clarification. I've been playing around with this, and I have a question:

How exactly do you select for Group Members whose Group Owner has self-cancelled and then passes the expiry date?

To select the Group Owner, you can use:
`if ('cancelled' === rcp_get_status() && rcp_is_expired() )`

But this doesn't work to select the Members in his group.

nosegraze:
`rcp_is_expired()` not working for group members is a bug that should be fixed here: #1061 That also affects the restricted content issue.

However I think `rcp_is_active()` should work as long as the member is a full member (and not still on `invited` status).

meadows19:
I've tried it and it looks like `rcp_is_active()` selects both (i) cancelled group members who _have_ passed their expiration date, and also (ii) cancelled group members who _have not_ passed their expiration date. So it can't be used to distinguish between both groups. Isn't is just pending and expired members who are not selected by `rcp_is_active()`?

Once the bug is fixed, this is what I think may work. I'll write it down here to help others who may search for the same issue.

**Select a normal user:** `if ((rcp_get_subscription_id($GLOBALS['user_ID']) == 1) { //code// }` Replace 1 with the subscription level ID.

**Select an expired user:** `if (rcp_is_expired()) { //code// }`

**Select a cancelled normal user before expiration, or cancelled Group Accounts Owner before expiration:** `if( 'cancelled' === rcp_get_status( $user_ID ) && ! rcp_is_expired( $user_ID )  { //code// }`

**Select a cancelled normal member or cancelled Group Accounts Owner after expiration:** `if( 'cancelled' === rcp_get_status( $user_ID ) &&  rcp_is_expired( $user_ID )  { //code// }` Since their status never gets changed to expired, even after the expiration date.

**Select a normal Group Accounts Member or Administrator:** ` if ( rcpga_group_accounts()->members->is_group_member() || rcpga_group_accounts()->members->is_group_admin() ) { //code// }`

**Select an expired Group Accounts Member or Administrator:** ` if ('expired' === rcp_get_status() && (rcpga_group_accounts()->members->is_group_member() || rcpga_group_accounts()->members->is_group_admin()) { //code// }`

These two currently don't work, but hopefully should work after `rcp_is_expired()` is made to apply to group account members too.

**Select a cancelled Group Accounts Member or Administrator before expiration:** `if ('cancelled' === rcp_get_status() && !rcp_is_expired() && (rcpga_group_accounts()->members->is_group_member() || rcpga_group_accounts()->members->is_group_admin()) { //code// }`

**Select a cancelled Group Accounts Member or Administrator after expiration:** `if ('cancelled' === rcp_get_status() && rcp_is_expired() && (rcpga_group_accounts()->members->is_group_member() || rcpga_group_accounts()->members->is_group_admin()) { //code// }`

Issue #999, opened 2017-02-20
If payment fails during free trial signup, user becomes ineligible to try again


nosegraze:
- User is eligible for trial and let's say they register via Stripe.
- Submit registration form.
- Their account is updated like so: `update_user_meta( $user_data['id'], 'rcp_has_trialed', 'yes' );`
- *Then* they're sent to the gateway class via `rcp_send_to_gateway()`
- Let's say the card is declined so they get an error screen.
- Go back to registration form to try again... but they can no longer sign up for a free trial since their account has already been flagged. So instead they're prompted to pay immediately.

The same would happen with PayPal if they get taken to PayPal but don't complete payment for whatever reason, then try again later.

Perhaps we should set some kind of `rcp_pending_trial_activation` meta instead, and only update to `rcp_has_trialed` and `rcp_is_trialing` after the account is activated? This would also fix a problem I'm having in #987

mindctrl:
A new pending meta key would work, but it highlights an issue with our payment flow. We really should not set the rcp_has_trialed flag until the member has truly initiated the trial with a successful payment.

nosegraze:
I feel like a lot of things would feel a bit cleaner if we perhaps created "pending" payment records immediately and then marked them as completed after payment is received, and with a bit more data in them (discount code used, etc.).

mindctrl:
That's definitely the direction we need to be moving. We have a need for pending payments in https://github.com/restrictcontentpro/restrict-content-pro/issues/798 and maybe https://github.com/restrictcontentpro/restrict-content-pro/issues/560.

pippinsplugins:
Pending payments would definitely help solve a _lot_ of issues with the registration flow.

pippinsplugins:
For now, how about just clearing out the meta on failure. Perhaps we could introduce a standard `rcp_registration_failed` hook that can be used to reset data?

nosegraze:
Is a failed registration something we can reliably check for in PayPal? If they're taken to the PayPal screen and then just close the window or something.

pippinsplugins:
If they click `Cancel and return to X`, we can control that URL.

nosegraze:
PR: #1016

Haven't done PayPal Standard yet.

We only want to delete the meta if the current signup was for a trial (to avoid deleting a flag that was possibly set like 2 signups ago or something). We could a query arg for the trial status in the `cancel_return` URL I guess, and delete the meta if that is present.

[...]

PR #1016, opened 2017-02-22
Delete trial meta if registration fails. #999

nosegraze:
#999

### DIFF
--- a/includes/registration-functions.php
+++ b/includes/registration-functions.php
@@ -943,7 +943,7 @@ add_action( 'rcp_set_status', 'rcp_remove_expiring_soon_email_sent_flag', 10, 2 
  */
 function rcp_remove_trial_flags_on_failure( $gateway ) {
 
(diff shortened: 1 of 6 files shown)


includes/registration-functions.php
```diff
@@ -932,3 +932,21 @@ function rcp_remove_expiring_soon_email_sent_flag( $status, $user_id ) {
...
+ * Remove trial flags if payment fails. This ensures they can try signing up for a trial again.
+ *
+ * @param RCP_Payment_Gateway $gateway
+ *
+ * @since  2.8
+ * @return void
+ */
+function rcp_remove_trial_flags_on_failure( $gateway ) {
+
+	if( $gateway->is_trial() ) {
```
pippinsplugins: Let's check that `$gateway->user_id` is not empty here, just to be extra safe.

nosegraze: Updated!

Issue #990, opened 2017-02-17
No information added to user when stripe charge fails


tnorthcutt:
Currently, if a charge in Stripe fails, the `rcp_stripe_charge_failed` action fires, but nothing is hooked to it, and no record is made of the failed charge in RCP.
```php
// failed payment
if ( $event->type == 'charge.failed' ) {
	do_action( 'rcp_stripe_charge_failed', $invoice );
	die( 'rcp_stripe_charge_failed action fired successfully' );
}
```

I propose that a note should be added that a charge failed. Here's a rough implementation:
```php
// failed payment
if ( $event->type == 'charge.failed' ) {
	do_action( 'rcp_stripe_charge_failed', $invoice );
        rcp_add_member_note( $member->ID, sprintf( __( 'Charge failed in Stripe. Event ID: %s', 'rcp' ), $event->id ) );
	die( 'rcp_stripe_charge_failed action fired successfully' );
}
```

mindctrl:
Good idea

nosegraze:
Maybe we should add them to all gateways via `rcp_recurring_payment_failed`.

Though then we wouldn't be able to include the `$event->id`

tnorthcutt:
@nosegraze yeah, would be really nice to have that event ID to make it easy to find the corresponding event in the Stripe admin interface.

nosegraze:
PR: #1022

I tried to do both. :P

tnorthcutt:
@nosegraze very nice! Leaves the option to add custom notes for other gateways in the future as well.

PR #1022, opened 2017-02-23
Add user note when recurring payment fails. #990

nosegraze:
#990

### DIFF
--- a/includes/member-functions.php
+++ b/includes/member-functions.php
@@ -1317,9 +1317,6 @@ add_action( 'rcp_set_status', 'rcp_update_expired_member_role', 10, 4 );
 /**
  * Add a note to the member when a recurring charge fails.
  *
- * Stripe and Stripe Checkout are handled separately to include the event ID.
- * @see rcp_add_stripe_recurring_payment_failure_note()
- *
  * @param RCP_Member          $member
  * @param RCP_Payment_Gateway $gateway
  *
--- a/includes/member-functions.php
+++ b/includes/member-functions.php
@@ -1285,3 +1285,23 @@ function rcp_update_expired_member_role( $status, $member_id, $old_status, $memb
 	}
 }
 add_action( 'rcp_set_status', 'rcp_update_expired_member_role', 10, 4 );
+
+/**
+ * Add a note to the member when a recurring charge fails.
+ *
+ * Stripe and Stripe Checkout are omitted from this because their notes are added directly
+ * in the gateway to include the event ID.
+ * @see RCP_Payment_Gateway_Stripe::process_webhooks()
+ *
+ * @param RCP_Member          $member
+ * @param RCP_Payment_Gateway $gateway
+ *
+ * @since 2.7.4
+ * @return void
+ */
+function rcp_add_recurring_payment_failure_note( $member, $gateway ) {
+	if ( 'RCP_Payment_Gateway_Stripe' != get_class( $gateway ) && 'RCP_Payment_Gateway_Stripe_Checkout' != get_class( $gateway ) ) {
+		rcp_add_member_note( $member->ID, __( 'Recurring charge failed.', 'rcp' ) );
+	}
+}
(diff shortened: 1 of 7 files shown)


includes/member-functions.php
```diff
@@ -1284,4 +1284,24 @@ function rcp_update_expired_member_role( $status, $member_id, $old_status, $memb
...
+ *
+ * @param RCP_Member          $member
+ * @param RCP_Payment_Gateway $gateway
+ *
+ * @since 2.7.4
+ * @return void
+ */
+function rcp_add_recurring_payment_failure_note( $member, $gateway ) {
+	if ( 'RCP_Payment_Gateway_Stripe' != get_class( $gateway ) && 'RCP_Payment_Gateway_Stripe_Checkout' != get_class( $gateway ) ) {
+		rcp_add_member_note( $member->ID, __( 'Recurring charge failed.', 'rcp' ) );
```
pippinsplugins: Could we move the note left for Stripe down here too? That way it's all in one place.

nosegraze:
Looks like `$event` was added to this hook in another PR:

`do_action( 'rcp_recurring_payment_failed', $member, $this, $event );`

Will that cause issues if we don't pass `$event` in any other gateway? Or can we just pass it in some but not others?

nosegraze:
Removed `$event` from `rcp_recurring_payment_failed`. It was added in https://github.com/restrictcontentpro/restrict-content-pro/pull/985 but I don't think it should be there in the non-Stripe-specific actions. It's still in `rcp_stripe_charge_failed` and can be used there.

tnorthcutt:
@nosegraze Since `$event` was only added in `class-rcp-payment-gateway-stripe.php`, it would only fire if Stripe is the acting gateway, right? Any downside to having it in there?

nosegraze:
Well `rcp_recurring_payment_failed` exists in all gateways. So I guess my worry was with mismatching parameter numbers if someone hooks into that for more than one gateway and it's expecting 2 parameters sometimes and 3 other times. I feel like I've seen warnings about that before but I'm not really sure.

Given that there's another action in there that _does_ include `$event` I think I'd rather omit it from `rcp_recurring_payment_failed` for consistency across all gateways.

mindctrl:
@nosegraze @pippinsplugins some questions and thoughts.

Why would Stripe have a standalone and not the others? As time goes on, we'll probably want to add the event ID in the note from other gateways, which would mean making a separate function for each if we follow this structure. Which means multiple functions fire and check for the proper gateway in every one of them, like the first line does in `rcp_add_recurring_payment_failure_note()`. Which gives us a bunch of gateway check code duplication just to add a payment failed note.

I'd love to see one function adding the note no matter the gateway used.

One idea is to save the relevant event ID, txn ID, whatever the gateway calls it, from the webhook as a `$this->webhook_event_id` variable before passing the gateway object to the hook. That would let one function add the note and reference the normalized variable name inside the object to get the event ID.

nosegraze:
> One idea is to save the relevant event ID, txn ID, whatever the gateway calls it, from the webhook as a $this->webhook_event_id variable before passing the gateway object to the hook.

I like that idea.

mindctrl:
Updated the PR with the idea mentioned above. Need to test the various gateways to make sure the proper transaction ID is being used.

nosegraze:
Looks good! Just added one small addition.

pippinsplugins:
I really like the way this has evolved :+1:

includes/member-functions.php
```diff
@@ -1312,4 +1312,32 @@ function rcp_update_expired_member_role( $status, $member_id, $old_status, $memb
...
 	}
 }
-add_action( 'rcp_set_status', 'rcp_update_expired_member_role', 10, 4 );
\ No newline at end of file
+add_action( 'rcp_set_status', 'rcp_update_expired_member_role', 10, 4 );
+
+/**
+ * Add a note to the member when a recurring charge fails.
+ *
+ * Stripe and Stripe Checkout are handled separately to include the event ID.
```
pippinsplugins: This comment needs removed since it has been standardized.

Issue #632, opened 2016-08-07
Add email verification


pippinsplugins:
Verification of email addresses during registration has been requested countless times. We really should add it.

Example request: https://secure.helpscout.net/conversation/236029190/39562/?folderId=841776

pippinsplugins:
Request: https://secure.helpscout.net/conversation/281754854/47454/?folderId=841781

nosegraze:
WIP: #1049

Just kind of beat it out, but it basically works. Let me know if you have any feedback.

To do:

- [ ] Add "Re-send verification email" link in _Restrict > Members_ interface when hovering over a member pending verification.
- [ ] Maybe add some kind of _"pending verification"_ note when editing the member as well.
- [ ] Allow subject/email to be customized in _Restrict > Settings > Emails_?
- [ ] Maybe filter the restricted content message if they're pending verification. We do this for WP Approve User.

pippinsplugins:
@nosegraze just left one comment on the PR.

Thoughts on putting this in a point release or making it one of the focus features for 2.9? I'm thinking point rrelease.

nosegraze:
Yep I'm happy with a point release!

mindctrl:
It would also make a nice key feature in 2.9. :)

nosegraze:
Ready for a final look and testing, I think!

I also changed the way notices are displayed on the edit profile shortcode.

pippinsplugins:
@nosegraze could we move the email verification settings to the top of the page, or perhaps just under the From name / address options?

pippinsplugins:
This works really well. One thing I think we need to add though, along with the adjustment of the settings placement.

The `Your account is pending email verification. Click here to re-send the verification email.` message should be shown at the top of [subscription_details]`. This is frequently placed on the registration success page and it can be jarring to get there and think your account is fully activated only to discover it's not when navigating elsewhere on the site.

nosegraze:
I moved the settings to the top, right above template tags. That work?

I'll work on moving the message to the other shortcode! 👍

pippinsplugins:
I think the message should be on both shortcodes, though it should also detect if both shortcodes are on the same page and only display once.

mindctrl:
This is looking good! Something I think we should clarify in the setting dropdown. `On for free registrations` could be mistaken for a registration that is free due to a 100% discount code. It might be clearer to say `On for free subscription levels` just to remove any doubts.

mindctrl:
[...]

Issue #1020, opened 2017-02-22
Option to require admin approval after registration, before granting access to content


nosegraze:
We currently integrate with WP Approve User to accomplish this, but we should build it directly into RCP instead.

1) People always complain about how old WP Approve User is (though it does still work).
2) WP Approve User requires that registration be enabled in Settings > General which is a bit annoying if you're using something like RCP and don't *actually* want that turned on.
3) It should probably be pretty easy to implement when we do this: https://github.com/restrictcontentpro/restrict-content-pro/issues/632 which is basically the same thing but with automatic approval after email confirmation, rather than manual admin approval. There could totally be some shared logic.

pippinsplugins:
Let's use the same user meta keys so that the transition between the two are seamless.

nosegraze:
Thinking about this some more because a few customers have had problems with WP Approve User. It requires that "anyone can register" be checked on, but a few users have conflicting plugins that are forcing that option off.

I definitely see the benefit of making the transition between WP Approve User and something built-in seamless, but I actually really dislike how WP Approve User does it. They do it the opposite way I would do it. Every single approved user gets a piece of user meta, and the unapproved users do not. This means whenever the plugin is activated, it needs to bulk add meta to every single existing user. I'd rather just have a piece of meta for the unapproved users and delete that meta when they become approved -- just like how we do email verification.

PR #1049, opened 2017-03-02
Email verification. #632

nosegraze:
#632


includes/class-rcp-member.php
```diff
@@ -1055,6 +1055,37 @@ public function has_trialed() {
...
 
 	/**
+	 * Determines if a member is pending email verification.
+	 *
+	 * @access public
+	 * @return bool
+	 */
+	public function is_pending_verification() {
+
+		$is_pending = get_user_meta( $this->ID, 'rcp_pending_email_verification', true );
```
pippinsplugins: Let's have this also look at the meta key set by the WP Approve User plugin so that they work seamlessly together.

nosegraze: So do you want this intertwined with WP Approve User (and eventually #1020) ? I guess I was seeing them as separate verification processes. Someone could theoretically require email verification _and_ admin approval, which might mean another method in #1020 for `is_pending_approval()`.

pippinsplugins: Hmmm, good question. I'm not sure right now actually. Will think on it.

includes/member-functions.php
```diff
@@ -1312,4 +1312,110 @@ function rcp_update_expired_member_role( $status, $member_id, $old_status, $memb
...
+		return false;
+	}
+
+	// The user should already be pending.
+	if ( ! rcp_is_pending_verification( $user_id ) ) {
+		return false;
+	}
+
+	$verify_link = add_query_arg( array(
+		'rcp-verify-key' => urlencode( $user->user_pass ),
```
mindctrl: Thoughts on using a unique key saved to meta temporarily instead of the hashed user pass? Theoretically more secure, and could be cleaned up during verification like the other keys.

nosegraze: Could save that unique key as the value in `rcp_pending_email_verification` meta.

includes/email-functions.php
```diff
@@ -445,6 +445,40 @@ function rcp_email_admin_on_manual_payment( $member, $payment_id, $gateway ) {
...
+	$message = isset( $rcp_options['verification_email'] ) ? $rcp_options['verification_email'] : '';
+	$message = apply_filters( 'rcp_verification_email', $message, $user_id );
+	$subject = isset( $rcp_options['verification_subject'] ) ? $rcp_options['verification_subject'] : __( 'Please confirm your email address', 'rcp' );
+	$subject = apply_filters( 'rcp_verification_subject', $subject, $user_id );
+
+	// @todo remove
+	//$message = sprintf( __( 'Click here to confirm your email address: %s', 'rcp' ), esc_url( rcp_generate_verification_link( $user_id ) ) );
+
+	error_log($message);error_log($subject);
+
```
mindctrl: We need to remove this.

nosegraze: Derp derp. :P

includes/class-rcp-member.php
mindctrl: I vote we let it works as is for now, as separate processes. If there's anything I'd change I think it would be if someone approves a member using WP Approve User, we go ahead and remove the pending email verification status... the thinking there being that if the admin thinks it's worth approving, it should just do it all around. Thoughts?

nosegraze: I like that plan!

Issue #1045, opened 2017-03-02
PayPal subscr_eot IPN incorrectly expiring account after new membership


nosegraze:
The `subscr_eot` IPN is triggered when a subscription plan reaches the end of its term. When we get that we expire the account:

```php
// user's subscription has reached the end of its term

if( 'cancelled' !== $member->get_status( $user_id ) ) {

	$member->set_status( 'expired' );

}
```

But it looks like this situation can cause incorrect expiration if a new plan was created. Example:

- User signs up for a recurring plan (Plan A).
- User cancels that plan with 2 weeks remaining.
- User signs up for a new plan (Plan B) and creates a new payment profile in PayPal.
- Then two weeks later, the `subscr_eot` is triggered for Plan A and this causes the account to be expired.

Haven't been able to replicate this myself, but it makes sense in theory. I didn't realize PayPal actually sent `subscr_eot` after a profile has been cancelled, but apparently they do.

pippinsplugins:
The subscription key should be included in the IPN. We could skip processing EOT IPNS if the subscription keys do not match.

nosegraze:
PR: #1070

PR #1070, opened 2017-03-09
Check payment profile ID before expiring member. #1045

nosegraze:
#1045


includes/gateways/class-rcp-payment-gateway-paypal.php
```diff
@@ -458,7 +458,7 @@ public function process_webhooks() {
 
 					// user's subscription has reached the end of its term
 
-					if( 'cancelled' !== $member->get_status( $user_id ) ) {
+					if( $posted['subscr_id'] == $member->get_payment_profile_id() && 'cancelled' !== $member->get_status() ) {
```
pippinsplugins: Let's add a sanity check here to make sure `subscr_id` is present.

nosegraze: Updated!

Issue #1060, opened 2017-03-07
PayPal Express: if initial payment fails, user still has access to content


nosegraze:
It looks like this flow is happening:

- User signs up for recurring payment.
- `recurring_payment_profile_created` IPN gets triggered and account is activated.
- Initial payment actually fails, triggering `recurring_payment_profile_cancel` IPN.
- RCP status is set to `cancelled`, but the account expiration date is still set in the future so they still retain access until the end of the cycle.

In the profile cancelled IPN, `initial_payment_status` comes back as `Failed` so that's something we could look for.

pippinsplugins:
I didn't know there was a `initial_payment_status` field. Cool.

nosegraze:
PR: #1074

mindctrl:
This looks good to me, but I can't test it right now. My sandbox is tripping again.

mindctrl:
The PR looks good, but I still can't get my sandbox working right to test. @nosegraze if you've confirmed this works, feel free to merge.

nosegraze:
Yep it works well. I turned on negative testing and made a purchase of `131.22` to trigger this response: `Transaction Declined (API Error Code 13122)`.

mindctrl:
Woot!

PR #1074, opened 2017-03-09
Set user back to pending if initial PayPal Express payment fails. #1060

nosegraze:
#1060

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-paypal-express.php
+++ b/includes/gateways/class-rcp-payment-gateway-paypal-express.php
@@ -522,7 +522,8 @@ class RCP_Payment_Gateway_PayPal_Express extends RCP_Payment_Gateway {
 						// Initial payment failed, so set the user back to pending.
 						$member->set_status( 'pending' );
 						$member->add_note( __( 'Initial payment failed in PayPal Express.', 'rcp' ) );
-						
+
+						do_action( 'rcp_registration_failed', $this );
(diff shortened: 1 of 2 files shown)


includes/gateways/class-rcp-payment-gateway-paypal-express.php
```diff
@@ -518,13 +518,19 @@ public function process_webhooks() {
 
 				if( ! $member->just_upgraded() ) {
 
-					// user is marked as cancelled but retains access until end of term
-					$member->cancel();
+					if( isset( $posted['initial_payment_status'] ) && 'Failed' == $posted['initial_payment_status'] ) {
+						// Initial payment failed, so set the user back to pending.
+						$member->set_status( 'pending' );
+						$member->add_note( __( 'Initial payment failed in PayPal Express.', 'rcp' ) );
```
pippinsplugins: Let's add an action here.

nosegraze: Added a new action `rcp_paypal_express_initial_payment_failed` but also `rcp_registration_failed`. Feels a bit weird to add that one in, since it's in a webhook rather than the initial registration, but maybe it actually makes sense because the initial signup would have failed? Thoughts?

mindctrl: I like having `rcp_registration_failed` too, since it's technically the initial signup and fits in with the rest.

Issue #1056, opened 2017-03-06
Upgrades improperly given both proration and full expiration date extension


nosegraze:
When a non-recurring member upgrades early, a proration credit is given and the expiration date is extended a full period from the current expiration date instead of current date. This results in the person getting free time on their subscription.

Related: #812

nosegraze:
PR: #1098

mindctrl:
If a non-recurring member renews the same subscription early, the remaining time is lost because the expiration date is calculated from today and not their existing date.

mindctrl:
Oh I think we fixed that already.

mindctrl:
We disabled proration in #812 when renewing same sub, but the expiration date is being calculated from now instead of existing expiration date.

nosegraze:
We fixed it for renewing the same sub, this one is specifically for upgrades.

PR #1098, opened 2017-03-14
Calculate from expiration today if prorating. #1056

nosegraze:
#1056

### DIFF
--- a/includes/registration-functions.php
+++ b/includes/registration-functions.php
@@ -195,8 +195,14 @@ function rcp_process_registration() {
 	// Delete pending expiration date in case a previous registration was never completed.
 	delete_user_meta( $user_data['id'], 'rcp_pending_expiration_date' );
 
+	// If they're given proration credits, calculate the expiration date from today.
+	$force_now = $auto_renew;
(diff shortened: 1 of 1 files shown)


includes/registration-functions.php
```diff
@@ -195,8 +195,14 @@ function rcp_process_registration() {
 	// Delete pending expiration date in case a previous registration was never completed.
 	delete_user_meta( $user_data['id'], 'rcp_pending_expiration_date' );
 
+	// If they're given proration credits, calculate the expiration date from today.
+	$force_now = $auto_renew;
+	if ( ! $force_now && ! empty( $member->get_prorate_credit_amount() ) ) {
```
pippinsplugins: Let's move `$member->get_prorate_credit_amount()` to a variable so it's outside of `empty()`. It's not required with PHP 5.3+ but it's good practice (in my opinion) to avoid it since using a function/method inside of `empty()` will fatal error on earlier versions.

nosegraze: I can't seem to stop putting functions in `empty()` !

Updated.

mindctrl: It's even worse! Prior to 5.5 you couldn't do functions inside `empty()`. Sad sad.

Issue #1100, opened 2017-03-14
Better RCP_Payment_Gateway documentation


nosegraze:
I'd love to add thorough docs for:

- [x] All properties.
- [x] All methods, explaining what they should do, maybe even with brief examples.
- [x] Link to corresponding documentation article in the header.

Almost like a mini version of the actual documentation article. It would be nice for a developer to be able to look at the class and thoroughly understand what's going on and how to extend the class.

nosegraze:
PR: #1111

PR #1111, opened 2017-03-20
Improve gateway documentation. #1100

nosegraze:
#1100

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway.php
+++ b/includes/gateways/class-rcp-payment-gateway.php
@@ -249,7 +249,7 @@ class RCP_Payment_Gateway {
 	 * the $this->initial_payment value. If recurring, you'll want to use $initial_payment for the
 	 * first payment and $this->amount for the recurring amount.
 	 *
-	 * After a successful registration, redirect to $this->checkout_url.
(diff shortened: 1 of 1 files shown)


includes/gateways/class-rcp-payment-gateway.php
```diff
@@ -59,32 +211,155 @@ public function __construct( $subscription_data = array() ) {
...
+	}
+
+	/**
+	 * Process registration
+	 *
+	 * This is where you process the actual payment. If non-recurring, you'll want to use
+	 * the $this->initial_payment value. If recurring, you'll want to use $initial_payment for the
+	 * first payment and $this->amount for the recurring amount.
+	 *
+	 * After a successful registration, redirect to $this->checkout_url.
```
mindctrl: I think this should be: After a successful payment, redirect to $this->return_url.

nosegraze: Good catch. No idea where I got `$this->checkout_url` from. :D

mindctrl: heh probably from the PayPal gateway, which does do that. Looks good, will merge once tests finish.

Issue #782, opened 2016-11-22
Restrict entire post type


nosegraze:
It would be nice to be able to restrict an entire custom post type with one setting. I've seen this request pop up a few times now.

Examples:
https://secure.helpscout.net/conversation/280575299/0/
https://secure.helpscout.net/conversation/278540313/46900/?folderId=841781

mindctrl:
I've seen a few requests for this too. The main thing that comes to mind is how it works with the existing restrictions UI at the post level. What about existing posts? What about ability to override at post level? When creating a new post, we'd need to read the default values for the post type and automatically set them on the post restriction settings... or introduce a new default option that flows back to the global post type setting. Also need to consider interaction with taxonomies.

nosegraze:
So we don't forget..

In the meeting we talked about the possibility of auto adding a new submenu page under each public post type that contains restriction settings for that post type.

nosegraze:
WIP PR: #1118

Certainly not done but I've started playing with some ideas. I haven't thought about overriding at a post level or anything right that. Honestly, I started cringing at how that would basically double the size of `RCP_Member::can_access()`. :P Right now it's basically:

- Whole post type is restricted
- Or, restrict individual posts

mindctrl:
> I haven't thought about overriding at a post level or anything right that. Honestly, I started cringing at how that would basically double the size of RCP_Member::can_access().

I agree with you. Anything that adds anything significant to `can_access()` gets a 👎 from me too.

nosegraze:
Maybe we should do something similar to what we did with content excerpts. If a post type is restricted then show all the settings in the restrict meta box but disable them, with a note about post type restriction.

nosegraze:
![rcp-restricted-post-type](https://cloud.githubusercontent.com/assets/6324272/24298265/5e421d7e-109d-11e7-9821-bd577d8daefc.png)

pippinsplugins:
Oh I love that.

nosegraze:
Ready for a look/testing!

pippinsplugins:
How about changing the `Restrict` submenu to `Restrictions` or `Restrict Access` or `Restrict Options`. Just `Restrict` doesn't seem clear enough to me.

pippinsplugins:
@nosegraze This is looking great! The restriction settings, however, aren't quite working for me.

[...]

PR #1118, opened 2017-03-21
Restricting an entire post type. #782

nosegraze:
#782


includes/admin/admin-pages.php
```diff
@@ -33,6 +33,13 @@ function rcp_settings_menu() {
 	$rcp_help_page          = add_submenu_page( 'rcp-members', __( 'Help', 'rcp' ), __( 'Help', 'rcp' ), 'rcp_view_help', 'rcp-help', '__return_null' );
 	$rcp_add_ons_page       = add_submenu_page( 'rcp-members', __( 'Add-ons', 'rcp' ), __( 'Add-ons', 'rcp' ), 'rcp_view_members', 'rcp-add-ons', 'rcp_add_ons_admin' );
 
+	// Add "Restrict" submenu under each post type.
+	foreach ( rcp_get_metabox_post_types() as $post_type ) {
+		$url  = ( 'post' == $post_type ) ? 'edit.php' : 'edit.php?post_type=' . $post_type;
+		$slug = ( 'post' == $post_type ) ? 'rcp-restrict' : 'rcp-restrict-' . $post_type;
+		add_submenu_page( $url, __( 'Restrict', 'rcp' ), __( 'Restrict', 'rcp' ), 'edit_posts', $slug, 'rcp_restrict_post_type' );
```
pippinsplugins: `edit_posts` may not always work here. Could we detect the edit capability on the post type and use that instead?

includes/admin/restrict-post-type-view.php
```diff
@@ -0,0 +1,105 @@
...
+				<?php foreach (
+					array(
+						'All',
+						'Administrator',
+						'Editor',
+						'Author',
+						'Contributor',
+						'Subscriber'
+					) as $role
+				) : ?>
```
mindctrl: This needs to support custom roles like we just added in #158.

includes/admin/restrict-post-type.php
```diff
@@ -0,0 +1,236 @@
...
+ * @since       2.9
+ */
+
+/**
+ * Render page for restricting an entire post type.
+ *
+ * @since 2.9
+ * @return void
+ */
+function rcp_restrict_post_type() {
```
mindctrl: This function's name doesn't accurate describe what it does.

mindctrl: I'd like to see this file and the view file both live in a subdirectory called `post-types`, keeping with the way we organize existing admin screens.

includes/admin/restrict-post-type.php
```diff
@@ -0,0 +1,236 @@
...
+	$url = add_query_arg( array(
+		'rcp_message' => 'post-type-updated'
+	), rcp_get_restrict_post_type_page( $post_type ) );
+
+	wp_safe_redirect( $url );
+	exit;
+
+}
+
+add_action( 'admin_init', 'rcp_save_post_type_restrictions' );
```
mindctrl: When we merge #1140, let's move this to the new admin action hook to avoid running this function at all when not needed.

includes/redirects.php
```diff
@@ -25,6 +25,17 @@ function rcp_redirect_from_premium_post() {
 				$redirect = home_url();
 			}
 			wp_redirect( $redirect ); exit;
+		} elseif( is_post_type_archive() && rcp_is_restricted_post_type( get_post_type() ) && ! $member->can_access( get_the_ID() ) ) {
```
mindctrl: `get_the_ID()` is `false` here, so this doesn't work as intended.

mindctrl: The global `$post` isn't available either. `get_queried_object()` will return the `WP_Post_Type` object, but that's not useful for `RCP_Member::can_access()`.

nosegraze: Hm how are you testing that? It's working just fine for me. It's returning the ID of the first post in the loop, which is what I expected.

mindctrl: @nosegraze here's what I did:

Enabled Hide Restricted Posts option, registered a new custom post type, restricted the whole post type, added a new post, saved it, then viewed the post type archive.

I tried two different **new** custom post types, not existing ones. Both fail to give a post ID.

Maybe I'm breaking something. Going to break and look at it again later.

nosegraze: Aha I think I've figured it out. Will look into it more.

nosegraze: Okay I haven't actually figured out _how_ this happened, but I know why you're seeing those results.

I started with a fresh post type and added just one post. I can't remember if I did that before or after I globally restricted the whole thing though...

But the problem is, I ended up with a single piece of RCP postmeta: `rcp_user_level` with a blank value. As a result, this one post was getting caught up in `rcp_get_restricted_post_ids()` and being removed from the query, meaning the post type archive had zero results, which of course meant `get_the_ID()` then didn't work. But deleting the empty `rcp_user_level` meta fixed the problem.

I've applied a patch that will account for this scenario, but I'd also like to figure out how the hell I ended up with that empty meta in the first place.

mindctrl: Yep! I had the blank `rcp_user_level` key too.

nosegraze: I think this commit should fix it if you want to give it another whirl: https://github.com/restrictcontentpro/restrict-content-pro/issues/1236 You'll need to create a new post and make sure the `rcp_user_level` isn't getting set.

includes/admin/post-types/restrict-post-type-view.php
```diff
@@ -0,0 +1,100 @@
...
+				<?php
+				$roles = get_editable_roles();
+				$roles = array_merge( array( 'all' => array( 'name' => 'All' ) ), $roles );
+				foreach(  $roles as $key => $role ) : ?>
+					<option value="<?php echo esc_attr( $key ); ?>"<?php selected( $key, strtolower( $user_role ) ); ?>><?php echo translate_user_role( $role['name'] ); ?></option>
+				<?php endforeach; ?>
+			</select>
+		</p>
+	</div>
+
```
pippinsplugins: We need to add an action here to allow extensions to register their own settings here. The actions in rcp_restrict_post_type_page() won't work since they are not inside of the `form` element.

includes/misc-functions.php
```diff
@@ -29,15 +29,29 @@ function rcp_is_sandbox(){
...
 		$post_id = get_the_ID();
+	}
 
 	$return = false;
+	$post_type_restrictions = rcp_get_post_type_restrictions( get_post_type( $post_id) );
+
+	if ( ! empty( $post_type_restrictions ) ) {
+
+		// Check post type restrictions.
+		if ( array_key_exists('is_paid', $post_type_restrictions ) ) {
```
pippinsplugins: Let's add a space for `'is_paid'` for formatting standards.

includes/redirects.php
```diff
@@ -25,6 +25,17 @@ function rcp_redirect_from_premium_post() {
 				$redirect = home_url();
 			}
 			wp_redirect( $redirect ); exit;
+		} elseif( is_post_type_archive() && $wp_query->have_posts() && rcp_is_restricted_post_type( get_post_type() ) && ! $member->can_access( get_the_ID() ) ) {
+			if( isset( $rcp_options['redirect_from_premium'] ) ) {
+				$redirect = get_permalink( $rcp_options['redirect_from_premium'] );
+			} else {
+				// Avoid a crazy redirect loop.
+				$redirect = ! is_home() ? home_url() : false;
```
pippinsplugins: This should be `is_front_page()` not `is_home()`. The `is_home()` function checks if we are on the blog page. Since `home_url()` returns the front page URL and not the blog page URL, this current implementation could result in a redirect loop if a member without access is on the frontpage and the blog page is different than the front page.

Issue #1095, opened 2017-03-13
Check subscription ID before cancelling membership via Stripe webhook


nosegraze:
It would be excellent if we could match the subscription ID before cancelling here:

```php
// Cancelled / failed subscription
if( $event->type == 'customer.subscription.deleted' ) {

	if( ! $member->just_upgraded() ) {

		$member->cancel();

		die( 'member cancelled successfully' );

	}

}
```

I've only had this happen once, but I ran into an issue while cancelling then immediately paying again:

- Cancel subscription.
- Manually adjust expiration date to be in the past.
- Register for new plan.
- Webhook triggers activating new plan.
- Slightly delayed webhook triggers cancelling the previous plan.
- Status ended up stuck at `cancelled`.

Again, I only had it happen once and it was probably just bad luck that the cancelled webhook was delayed, but would still be nice to fix just in case!

nosegraze:
PR: #1130

pippinsplugins:
PR looks good to me.

PR #1130, opened 2017-03-23
Only cancel subscription if IDs match. #1095

nosegraze:
#1095


includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -581,7 +581,7 @@ public function process_webhooks() {
 					// Cancelled / failed subscription
 					if( $event->type == 'customer.subscription.deleted' ) {
 
-						if( ! $member->just_upgraded() ) {
+						if( ! $member->just_upgraded() && $payment_event->id == $member->get_merchant_subscription_id() ) {
```
mindctrl: Haven't tested, but couldn't we get rid of the `just_upgraded()` check now that we're matching the ID?

nosegraze: Yeah you're probably right!

Issue #1131, opened 2017-03-24
Only show admin notices to people who have necessary capabilities


nosegraze:
Our admin notices are visible to all logged in users, even subscribers. Particularly:

- Notice to configure RCP settings.
- Database needs an upgrade.
- Expired license key.
- License key not entered.
- Outdated integrations.

Also all the action based ones ("Subscription level updated") but it's unlikely an unauthorized person would run into those (but we could still fix them).

nosegraze:
PR: #1132

PR #1132, opened 2017-03-24
Check capabilities before showing admin notices #1131

nosegraze:
#1131

### DIFF
--- a/includes/admin/admin-notices.php
+++ b/includes/admin/admin-notices.php
@@ -20,149 +20,174 @@ function rcp_admin_notices() {
 	$class   = 'updated';
 	$text    = '';
 
-	// only show notice if settings have never been saved
-	if( ! is_array( $rcp_options ) || empty( $rcp_options ) ) {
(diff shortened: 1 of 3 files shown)


includes/admin/admin-notices.php
```diff
@@ -20,149 +20,174 @@ function rcp_admin_notices() {
...
-		echo '<div class="updated fade"><p>' . __( 'The Restrict Content Pro database has been updated', 'rcp' ) . '</p></div>';
-	}
-
-	if( false !== get_transient( 'rcp_login_redirect_invalid' ) ) {
-		echo '<div class="error"><p>' . __( 'The page selected for log in redirect does not appear to contain a log in form. Please add [login_form] to the page then re-enable the log in redirect option.', 'rcp' ) . '</p></div>';
-	}
-
-	if ( 'expired' === rcp_check_license() && ! get_user_meta( get_current_user_id(), '_rcp_expired_license_dismissed', true ) ) {
-		echo '<div class="error info">';
+	if( current_user_can( 'rcp_manage_settings' ) ) {
```
pippinsplugins: Instead of wrapping the notices in the check, how about just bailing early? That way we don't have everything indented and it's easier to not mess up in the future by adding a new notice after the closing `}`.

pippinsplugins: Oh wait, nevermind, just realized you're doing multiple checks depending on the notices. All good :+1:

Issue #1103, opened 2017-03-16
Add debugging class


mindctrl:
Let's add an option to enable debugging to a custom log file. There's one in AffWP we can use.

nosegraze:
PR: #1133

I only added one log so far so I could test it. To do:

- [ ] Add more logs.
- [ ] Replace uses of `WP_Logging`.
- [ ] Remove `WP_Logging` class.

One thing to consider: we talked about how the file is publicly accessible, but it's fine because it only contains anonymous data. But then we also discussed removing the "Email IPN report" in favour of this new logging system.. but if we integrate IPN reports then that would contain data like email addresses.

mindctrl:
I wouldn't want us to log any personal data to a publicly accessible file. Given that, I think the IPN option should stay the same.

nosegraze:
Updated PR:

- Added more logs.
- Also added more actions in the gateways for specific events. A lot of the existing actions were gateway-specific, i.e. `rcp_authorizenet_silent_post_payment`. Now they all share one like `rcp_gateway_payment_processed`.
- Removed old "Logs" menu item and deprecated WP_Logging class. I figured maybe we should keep the actual class there for a while in case people are using it in custom gateways?

I'll add logs for admin actions once #1101 is merged.

Here's a sample log for a single new registration:

```
2017-4-14 14:12:51 - Started new registration for subscription #15 via stripe.
2017-4-14 14:12:51 - Registration for user #1 sent to gateway. Level ID: 15; Initial Amount: 50; Recurring Amount: 50; Auto Renew: false
2017-4-14 14:12:55 - New payment inserted. ID: 106; User ID: 1; Amount: 50; Subscription: 1 Year Plan; Status: complete
2017-4-14 14:12:58 - Payment Received email sent to user #1.
2017-4-14 14:12:58 - New payment #106 inserted for member #1 via Stripe gateway.
2017-4-14 14:12:58 - Updating recurring status for member #1. Previous: false; New: false
2017-4-14 14:13:00 - Active email sent to user #1.
2017-4-14 14:13:02 - Active email sent to admin(s).
```

mindctrl:
Adjustments to the constructor look good. I was able to whip up a quick test and do a custom log.

---

Seeing a couple log issues. One is really a symptom of a different issue, but worth mentioning.

[...]

PR #1133, opened 2017-03-27
Add debugging to log file. #1103

nosegraze:
#1103


includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -417,6 +417,8 @@ protected function handle_processing_error( $e ) {
 		$error .= "<p>Status: " . $e->getHttpStatus() ."</p>";
 		$error .= "<p>Message: " . $err['message'] . "</p>";
 
+		rcp_log( sprintf( 'Stripe registration failure for user #%d: %s', $this->user_id, $err['message'] ) );
```
mindctrl: I'd rather add the log entry from another file, via one of the failed action hooks above this line. I want to start moving all non-payment related functionality out of the gateway classes.

nosegraze: Good plan. Mostly just threw in something random to make sure it actually worked. :P

nosegraze: Moved it to a hook and added a new `$error_message;` property so we can still get the error message. Like that approach?

mindctrl: Yes, that looks good.

includes/class-rcp-logging.php
```diff
@@ -0,0 +1,157 @@
...
+	 * @since  2.9
+	 * @return void
+	 */
+	public function clear_log() {
+		@unlink( $this->file );
+	}
+
+}
+
+$GLOBALS['rcp_logs'] = new RCP_Logging();
```
mindctrl: Can we instantiate inside `rcp_log()` instead? That way we can ditch the global and only instantiate the class when it's used.

mindctrl: Only disadvantage to that I can think of would be if you need to log multiple items from separate places in a single load, but that's maybe improbable?

nosegraze: I had the same thought about logging multiple items, which is why I did it this way. I hate globals though so I'm open to whatever.

mindctrl: 👍 let's move it and if it turns out that it's a performance issue due to spinning up and tearing down the class numerous times, we can change it.

includes/class-rcp-logging.php
```diff
@@ -0,0 +1,155 @@
...
+	 * @access public
+	 * @since  2.9
+	 * @return void
+	 */
+	public function init() {
+
+		$upload_dir     = wp_upload_dir();
+		$this->filename = 'rcp-debug.log';
+		$this->file     = trailingslashit( $upload_dir['basedir'] ) . $this->filename;
+
```
mindctrl: I'd like these values passed into the class constructor, maybe an an array with these default values parsed with `wp_parse_args()`, which would make it viable for other logging use cases on the fly and not hardcoded for our specific use case. Also, with that there won't be any need for the `init()` method. We can set the values directly in `__construct()`.

includes/class-rcp-logging.php
```diff
@@ -0,0 +1,155 @@
...
+		if ( ! is_writeable( $upload_dir['basedir'] ) ) {
+			$this->is_writable = false;
+		}
+
```
mindctrl: Since things could change between class instantiation and the time a log file is written, we should probably check `is_writeable` right before the write operation and not here. Also, we're not really using `$this->is_writable` anywhere in the class before any write operation. We're using it in `get_file()`, but the use doesn't make sense and it doesn't check before writing the empty log file.

includes/class-rcp-logging.php
```diff
@@ -0,0 +1,155 @@
...
+		} else {
+
+			@file_put_contents( $this->file, '' );
+			@chmod( $this->file, 0664 );
+
+		}
+
+		return $file;
+	}
```
mindctrl: This method should do one thing, get the file. It's writing an empty file here if one doesn't already exist, but I don't think we need to do more than one thing.

includes/class-rcp-logging.php
```diff
@@ -0,0 +1,155 @@
...
+	 * @since  2.9
+	 * @return void
+	 */
+	protected function write_to_log( $message = '' ) {
+
+		$file = $this->get_file();
+		$file .= $message;
+		@file_put_contents( $this->file, $file );
+
+	}
```
mindctrl: This should check that the file is writable before trying to write it.

includes/class-rcp-logging.php
```diff
@@ -0,0 +1,155 @@
...
+	 * Clear the log
+	 *
+	 * @access public
+	 * @since  2.9
+	 * @return void
+	 */
+	public function clear_log() {
+		@unlink( $this->file );
+	}
+
```
mindctrl: This should check that the file is writable before trying to delete it. I think generally ability to delete is the same as the ability to write.

Issue #1148, opened 2017-03-31
Incorrectly matching term_taxonomy_id with term_id when checking for term restricted posts


nosegraze:
```php
$p_ids = $wpdb->get_results( $wpdb->prepare( "SELECT object_id FROM {$wpdb->term_relationships} WHERE term_taxonomy_id = %d", absint( $term->term_id ) ), ARRAY_A );
```

We're matching `term_taxonomy_id` against the `term_id` - two different things. We should be using `$term->term_taxonomy_id` instead.

nosegraze:
This function has a similar problem: `rcp_get_connected_term_ids()`

It's designed to return term IDs but it's actually returning the `term_taxonomy_id`.

nosegraze:
PR: #1149

PR #1149, opened 2017-03-31
Fix incorrect usage of term_taxonomy_id. #1148

nosegraze:
#1148


nosegraze:
Damn, `object_ids` wasn't added until 4.7.

Issue #847, opened 2017-01-08
Add status filters to Subscription Levels, Discount Codes, and Payments tables


ramiy:
Sites with many subscription levels have a very messy subscription screen.

It would be nice to have some kind of indication whether the subscription levels are active or inactive.

We can add a new column with Red/Green color to the table.

Another approach is to add a navigation filter above the table:

![nav](https://cloud.githubusercontent.com/assets/576623/21747324/22aebe14-d56c-11e6-82e9-379f95c4fe38.png)

ramiy:
By the way, the discounts table has a "**status**" column. The same column can be added to the subscription levels table.

ramiy:
Status filter are available only on **Members** page:

![rsc-members](https://cloud.githubusercontent.com/assets/576623/22011406/99713e64-dc97-11e6-8c2d-87fd3cc768a7.png)

It would be nice to have status filters on:
* **Subscription Levels** Screen
* **Discount Codes** Screen
* **Payments** Screen

nosegraze:
PR: #1170

I added status filters to the Subscription Levels, Discount Codes, and Payments tables.

I also ended up tweaking some of the formatting of the payments DB queries, mostly to ensure everything is properly escaped and running through `$wpdb->prepare()` whenever possible. That probably ended up being a bit out of the scope of this issue, but I got carried away!

pippinsplugins:
@nosegraze Looking good!

Seems that payment counts aren't quite working:
<img width="762" alt="screen shot 2017-04-12 at 11 58 59 am" src="https://cloud.githubusercontent.com/assets/1034109/24969763/7a1f85de-1f77-11e7-89f0-1d2fe6cc527c.png">
I refunded a payment and it shows in refunded but the count is still 0.

nosegraze:
Maybe the transient didn't clear properly. I'll take a look!

mindctrl:
This is a minor thing, but if you view a status with 0 discount codes, the table says "No discount codes added yet".

![image](https://cloud.githubusercontent.com/assets/1254870/25345270/a189c8f4-28e2-11e7-95f0-9dd0e1857a4c.png)

mindctrl:
We could not show the link if the count = 0. This is what WP core does.

mindctrl:
Updated a payment from refunded to complete and the refunded count didn't update.

[...]

PR #1170, opened 2017-04-12
Add status filters to Subscription Levels, Discount Codes, and Payments tables #847

nosegraze:
#847


includes/admin/discounts/discount-codes.php
```diff
@@ -14,15 +14,45 @@
...
  */
 function rcp_discounts_page() {
-	global $rcp_options, $rcp_discounts_db_name, $wpdb;
-	$page = admin_url( '/admin.php?page=rcp-discounts' );
+	/**
+	 * @var RCP_Discounts $rcp_discounts_db
+	 */
+	global $rcp_discounts_db;
+	$page   = admin_url( '/admin.php?page=rcp-discounts' );
+	$status = isset( $_GET['status'] ) ? $_GET['status'] : 'all';
```
pippinsplugins: Let's sanitize `$_GET['status']` please

includes/admin/payments/payments-page.php
```diff
@@ -37,8 +37,8 @@ function rcp_payments_page() {
 
 		$rcp_payments  = new RCP_Payments();
 		$page          = isset( $_GET['p'] ) ? $_GET['p'] : 1;
-		$per_page      = 20;
 		$search        = ! empty( $_GET['s'] )       ? urldecode( $_GET['s'] )      : '';
+		$status        = isset( $_GET['status'] ) ? $_GET['status'] : '';
```
pippinsplugins: `$_GET['status']` should be sanitized.

includes/admin/subscriptions/subscription-levels.php
```diff
@@ -14,39 +14,68 @@
  * @return void
  */
 function rcp_member_levels_page() {
-	global $rcp_options, $rcp_db_name, $wpdb;
-	$page = admin_url( '/admin.php?page=rcp-member-levels' );
+	global $rcp_options, $rcp_db_name, $wpdb, $rcp_levels_db;
+	$page   = admin_url( '/admin.php?page=rcp-member-levels' );
+	$status = isset( $_GET['status'] ) ? $_GET['status'] : 'all';
```
pippinsplugins: `$_GET['status']` should be sanitized.

includes/admin/discounts/discount-codes.php
```diff
@@ -14,15 +14,45 @@
...
+				</li>
+				<li>
+					<a href="<?php echo esc_url( add_query_arg( 'status', 'active', $page ) ); ?>" title="<?php esc_attr_e( 'View active discount codes', 'rcp' ); ?>"<?php echo 'active' == $status ? ' class="current"' : ''; ?>>
+						<?php _e( 'Active', 'rcp' ); ?>
+						<span class="count">(<?php echo $active_count; ?>)</span>
+					</a>|
+				</li>
+				<li>
+					<a href="<?php echo esc_url( add_query_arg( 'status', 'disabled', $page ) ); ?>" title="<?php esc_attr_e( 'View inactive discount codes', 'rcp' ); ?>"<?php echo 'disabled' == $status ? ' class="current"' : ''; ?>>
+						<?php _e( 'Inctive', 'rcp' ); ?>
```
mindctrl: Misspelled `Inactive`.

includes/admin/subscriptions/subscription-levels.php
```diff
@@ -14,39 +14,68 @@
...
+				</li>
+				<li>
+					<a href="<?php echo esc_url( add_query_arg( 'status', 'active', $page ) ); ?>" title="<?php esc_attr_e( 'View active subscription levels', 'rcp' ); ?>"<?php echo 'active' == $status ? ' class="current"' : ''; ?>>
+						<?php _e( 'Active', 'rcp' ); ?>
+						<span class="count">(<?php echo $active_count; ?>)</span>
+					</a>|
+				</li>
+				<li>
+					<a href="<?php echo esc_url( add_query_arg( 'status', 'inactive', $page ) ); ?>" title="<?php esc_attr_e( 'View inactive subscription levels', 'rcp' ); ?>"<?php echo 'inactive' == $status ? ' class="current"' : ''; ?>>
+						<?php _e( 'Inctive', 'rcp' ); ?>
```
mindctrl: Same misspelling here.

includes/class-rcp-payments.php
```diff
@@ -243,41 +251,39 @@ public function get_payments( $args = array() ) {
...
+			if( is_array( $args['status'] ) ) {
+				$status_count       = count( $args['status'] );
+				$status_placeholder = array_fill( 0, $status_count, '%s' );
+				$statuses           = implode( ', ', $status_placeholder );
+
+				$where .= " AND `status` IN( {$statuses} ) ";
 
-			if( ! empty( $args['subscription'] ) || ! empty( $args['user_id'] ) ) {
-				$where .= "AND `status` IN( {$statuss} ) ";
+				$where .= $wpdb->prepare( "AND `status` IN ( %s )", $statuses );
```
mindctrl: This duplicates the line above, resulting in an invalid WHERE clause.
Example: ```WHERE 1=1  AND `status` IN( pending, refunded ) AND `status` IN ( 'pending, refunded' )```

Also, as far as I can tell, getting an array of statuses doesn't work and I'm not sure it ever has. Definitely needs to be fixed.

mindctrl: Also I think the placeholder/status bit is wrong, which I changed locally in testing. Otherwise the values don't get populated properly.

mindctrl: By that I mean, it ends up being: `WHERE 1=1  AND `status` IN( %s, %s ) AND `status` IN ( '%s, %s' )`

nosegraze: Updated. Did a quick test on my end and I think that fixes it.

includes/class-rcp-discounts.php
```diff
@@ -61,18 +61,62 @@ public function get_discounts( $args = array() ) {
 
 		$where = '';
 
+		// Filter by status.
+		if ( $args['status'] == 'active' ) {
+			$where = "WHERE `status` != 'disabled'";
+		} elseif ( $args['status'] == 'disabled' ) {
+			$where = "WHERE `status` = 'disabled'";
+		}
+
```
mindctrl: Need to check that `$args[status]` isn't empty like before. WIth that, it results in PHP notices, and since `rcp_get_discounts()` used to take no args, that's a likely scenario.

includes/admin/payments/payments-page.php
```diff
@@ -37,8 +37,8 @@ function rcp_payments_page() {
 
 		$rcp_payments  = new RCP_Payments();
 		$page          = isset( $_GET['p'] ) ? $_GET['p'] : 1;
```
mindctrl: We should probably `absint` `$_GET['p']` too while we're here.

includes/class-rcp-discounts.php
mindctrl: We could also run a default value through `wp_parse_args` like the `count()` method does. Either would be fine I think.

includes/class-rcp-levels.php
```diff
@@ -164,6 +164,45 @@ public function get_levels( $args = array() ) {
...
+			'status' => 'all'
+		);
+
+		$args = wp_parse_args( $args, $defaults );
+
+		$where = '';
+
+		// Filter by status.
+		if ( $args['status'] == 'active' ) {
+			$where = "WHERE `status` !='inactive'";
```
mindctrl: I know this is the same as `get_levels()` above, but I think we should move both to:
```$where = "WHERE `status` !='active'";```

unless there's a reason not to. Technically it's incorrect to count/return all statuses except `inactive` when someone requests `active`. The `insert` and `update` methods allow for custom statuses already, so I don't see any reason we shouldn't account for it here and in `get_levels()`.

Comment (mindctrl) [inline includes/class-rcp-levels.php]: Oops, edited my comment due to an error in the suggestion.

File: includes/class-rcp-payments.php [hunk 10]
```diff
@@ -388,49 +384,81 @@ public function count( $args = array() ) {
...
+
+		// Search
+		if( ! empty( $args['s'] ) ) {
+
+			// Search by email
+			if( is_email( $args['s'] ) ) {
+
+				$user = get_user_by( 'email', $args['s'] );
+
+				$where .= $wpdb->prepare( " AND `user_id` = %d", $user->ID );
```
mindctrl: This throws a notice if an invalid user email is supplied.

includes/class-rcp-payments.php
```diff
@@ -314,19 +314,15 @@ public function get_payments( $args = array() ) {
...
-			else
-				$where = " AND ";
-
 			// Search by email
 			if( is_email( $args['s'] ) ) {
 
 				$user = get_user_by( 'email', $args['s'] );
 
-				$where .= "`user_id`=$user->ID ";
+				$where .= $wpdb->prepare( " AND `user_id` = %d", $user->ID );
```
mindctrl: This throws a notice if an invalid user email is supplied.

includes/class-rcp-payments.php
nosegraze: Fixed both. https://github.com/restrictcontentpro/restrict-content-pro/pull/1170/commits/cf4a250b88fc771d0f705cfc373cb684e70e99da

Issue #765, opened 2016-11-12
Option to auto add new users to subscription level


nosegraze:
> "You should think of having this as a setting in RCP so new users get defaulted to a specified membership. Plugins like MyMail do this for newsletter subscribers."

Ref: https://secure.helpscout.net/conversation/274782280/46322/?folderId=1100052

Could be useful for free subscription levels.

nosegraze:
Started here: #1190

Can finish it off after #713. That function would be handy.

mindctrl:
I think we need to clarify the description here. It's not clear on what this does and does not do. "New WordPress users" isn't very clear, and a lot of our customers don't even know you can register new users at /wp-login.php or otherwise.

I think we should expand on this and explain that it does not apply to user accounts created during subscription registration, and that it only applies to user accounts created via other means.

mindctrl:
Also to avoid confusion, we need to make it clear that this does not add a recurring subscription. That's obvious to us, but I've had customers think manually adding a subscription level to a user account in the admin magically created a payment profile for the member........

nosegraze:
How about:

>  Check this to automatically add new WordPress users to a subscription level. This only needs to be turned on if you're adding users manually or through some means other than the registration form. This does not automatically take payment so it's best used for free levels.

Or if that's too wordy we could cut it down and link to a documentation article.

mindctrl:
I think that's fine. If we get any confusion about it we can adjust it.

pippinsplugins:
I'm good with that phrasing.

nosegraze:
This is ready for testing now.

mindctrl:
I think the `rcp_auto_register_user()` function should be renamed. It's not auto registering user accounts, it's adding subscriptions to accounts that have been registered via other means.

nosegraze:
Hmm.. `rcp_auto_add_subscription_level_to_user()` ? Or `rcp_auto_grant_membership()` ? Any other ideas?

mindctrl:
I renamed it to `rcp_user_register_add_subscription_level()`. Thoughts on that?

This is tested and works well. Once we settle on the name we're good to merge.

nosegraze:
Fine with me!

pippinsplugins:
Looks good to me.

PR #1190, opened 2017-04-19
WIP: Auto add new users to subscription level. #765

nosegraze:
#765


includes/registration-functions.php
```diff
@@ -1228,3 +1228,39 @@ function rcp_add_subscription_to_user( $user_id, $args = array() ) {
...
+	if ( empty( $level_id ) ) {
+		return;
+	}
+
+	// Don't run if we're on the registration form.
+	if ( did_action( 'rcp_form_errors' ) ) {
+		return;
+	}
+
+	rcp_add_subscription_to_user( $user_id, array(
```
pippinsplugins: I know this function came from a different PR but can we make a small change?

It bugs me that it's "add subscription to user" yet the first parameter in the function is the user ID.

Can we rename it to `rcp_add_user_to_subscription()`?

nosegraze: I'm fine with that change. I'll make it now!

Issue #1197, opened 2017-04-21
Submit button disabled if reCaptcha is checked on but no keys are filled out


nosegraze:
Really this is a user error, but maybe we should account for it just in case.

We pass this into the JavaScript:

```php
'recaptcha_enabled'  => isset( $rcp_options['enable_recaptcha'] ) ? true : false
```

Then the JS does this:

```js
if ( '1' === rcp_script_options.recaptcha_enabled ) {
	jQuery('#rcp_registration_form #rcp_submit').prop('disabled', true);
}
```

If reCaptcha is checked on but no keys are entered, the reCaptcha doesn't actually get displayed on the registration form, but the submit button gets disabled anyway.

mindctrl:
I've seen this twice already. People are amazing. Enable it but don't enter the keys.

nosegraze:
PR: #1198

PR #1198, opened 2017-04-21
Introduce function checking if reCAPTCHA is enabled. #1197

nosegraze:
#1197

### DIFF
--- a/includes/captcha-functions.php
+++ b/includes/captcha-functions.php
@@ -10,6 +10,21 @@
  */
 
 /**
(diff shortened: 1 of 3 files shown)


mindctrl:
This is because I have recaptcha disabled and the file only loads if it's enabled.

mindctrl:
Secondary note, just for consistency I think the function should be named `rcp_is_recaptcha_enabled()`

nosegraze:
Updated.

I changed it to always include the `captcha-functions.php` file. I don't think there's any harm in that since we were always checking if it was enabled before running any of the code anyway.

Issue #1213, opened 2017-04-26
WooCommerce: products in restricted category/tag are still accessible via direct URL


nosegraze:
They're hidden from the shop archive, but you can still access them via a direct URL.

nosegraze:
PR: #1214

PR #1214, opened 2017-04-26
Restrict direct access to term restricted products #1213

nosegraze:
#1213

### DIFF
--- a/includes/integrations/class-rcp-woocommerce.php
+++ b/includes/integrations/class-rcp-woocommerce.php
@@ -374,11 +374,11 @@ class RCP_WooCommerce {
 		$tag_restricted = true === $product_tag ? true : false;
 
 		// Return early if no restrictions
-		if ( ! $active_only && empty( $levels[0] ) && ! $access_level && ! $cat_restricted && ! $tag_restricted ) {
+		if ( ! $active_only && empty( $levels ) && ! $access_level && ! $cat_restricted && ! $tag_restricted ) {
 			return $template;
 		}
 
-		$visible = true;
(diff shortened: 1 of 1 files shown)


includes/integrations/class-rcp-woocommerce.php
```diff
@@ -374,11 +374,11 @@ public function hide_template( $template, $slug, $name ) {
 		$tag_restricted = true === $product_tag ? true : false;
 
 		// Return early if no restrictions
-		if ( ! $active_only && empty( $levels[0] ) && ! $access_level && ! $cat_restricted && ! $tag_restricted ) {
+		if ( ! $active_only && empty( $levels ) && ! $access_level && ! $cat_restricted && ! $tag_restricted ) {
```
mindctrl: If memory serves me correctly, this originally checked `$levels[0]` because of some legacy data. If `$levels` looks like this:
```
array( 0 => array() );
```
then just doing `empty( $levels )` will return false.

There are also some other instances in this file that check `$levels[0]`, which makes me think the legacy thing I'm remembering is right. With the legacy thing fixed, `$levels` will be an empty string, instead of an array, if there are no restrictions defined. Luckily `empty()` won't complain if we check `empty( $levels[0] )` on an empty string.

nosegraze: Should we clean up that legacy data? Seems a bit gross to be saving `array( 0 => array() );`

mindctrl: It looks like we don't save it that way now. We can clean them up later when we have batch processing.

Issue #1230, opened 2017-05-10
Update registration js to pass all form fields


tannerm:
Update rcp_calc_total() function to pass all form values in the ajax call like the form submission action.

PR #1231, opened 2017-05-10
Update rcp_calc_total to pass all form values in the ajax call

tannerm:
#1230


includes/js/register.js
```diff
@@ -373,14 +374,11 @@ function rcp_calc_total() {
...
+		rcp_gateway: $gateway.val(),
+		rcp_form_data: form.serialize()
 	};
 
-	if ( $('#rcp_auto_renew').is(':checked') ) {
-		data.rcp_auto_renew = true;
-	}
-
-	$.post(rcp_script_options.ajaxurl, data, function(response) {
+	$.post( rcp_script_options.ajaxurl, form.serialize() + '&action=rcp_calc_discount', function(response) {
```
mindctrl: The form is serialized in `data` above, but not used here. We should do one or the other.

Also, I think it's probably best not to send the password fields every time a calculation is triggered.

mindctrl: Also this would need to be based off and merged into `release/2.9`.

mindctrl: Sticking with the `data` var would make it easier for us to add any custom data attributes to the post request, such as `data-blah-blah` attributes that we use in some places.

tannerm: Good thought @mindctrl! I just pushed the change and added to the excluded list 'rcp_card_number' and 'rcp_register_nonce' (since this was triggering registration).

tannerm:
Created new pull request from master at #1240

Issue #1284, opened 2017-06-15
Composer/Installers version incompatible with Bedrock


fourstacks:
I'm trying to install RCP via composer in a Bedrock project.  The version of composer/installers in composer.json is set at 1.0.6 and the version included with bedrock is at 1.2.0 so composer throws a compatibility error.  Proposing a PR to fix this

tnorthcutt:
Closing this as the related PR is also closed: https://github.com/restrictcontentpro/restrict-content-pro/pull/1285

PR #1285, opened 2017-06-15
Bump composer/installers version #1284

fourstacks:
Current composer/installers version number in composer.json causes version conflict when used with Bedrock. Increased to v1.3.0 for compatibility.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,6 @@
 		}
 	],
 	"require": {
-		"composer/installers": "~1.3.0"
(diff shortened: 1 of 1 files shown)


mindctrl:
@fourstacks sorry for missing this. It looks like you reverted the version from 1.3.x to 1.0.x. Is this the way you intended it to be?

mindctrl:
@fourstacks 👍  thanks!

PR #1311, opened 2017-07-11
Don't override join date during registration if it's already set #1286

### LINKED ISSUE #1286: Don't override join date during registration if it's already set
During registration, the join date is _always_ set, without checking if it already exists. So if a user joins May 10th, their join date is May 10th. Then the user renews manually June 10th and their join date is updated to June 10th.

A customer reported this issue because it's interfering with Drip Content. When the user renews, their drip schedule resets and they actually lose access to content they already had previously.

dbeja:
Fixes #1286.

Basically, first I check if the joined_date for that subscription ID already exists, if it exists I don't update the date.

I only have 2 questions:
- If the joined_date for that subscription already exists I'm returning true. Should I return false instead?
- Also, if the joined_date already exists I'm still calling the action rcp_set_joined_date. Should I leave it?

### DIFF
--- a/includes/class-rcp-member.php
+++ b/includes/class-rcp-member.php
@@ -289,9 +289,10 @@ class RCP_Member extends WP_User {
 		}
 
 		// check if joined date already exists.
-		$has_joined_date = ! empty( get_user_meta( $this->ID, 'rcp_joined_date_' . $this->get_subscription_id(), true ) );
+		$joined_date = get_user_meta( $this->ID, 'rcp_joined_date_' . $this->get_subscription_id(), true );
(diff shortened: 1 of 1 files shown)


nosegraze:
Thanks for the PR @dbeja :)

I actually think we should leave the `RCP_Member::set_joined_date()` method as-is so people are able to override the join date if they need to for whatever reason. Instead, I'd rather add the `if` statement in the registration function. However, we'll probably want to hold off on making a PR for that because the registration function will change quite a bit in the `2.9` update so it'd be better to wait until that's merged to avoid some gnarly merge conflicts.

dbeja:
I understand!
Thanks for the feedback!

Issue #1304, opened 2017-07-03
Add usage tracking


nosegraze:
We've talked about this in our meetings. This will be useful for determining which add-ons are the most commonly used.

See https://github.com/AffiliateWP/AffiliateWP/issues/1498

nosegraze:
What RCP data do we want to collect? (Other than the standard stuff like PHP version, active plugins, etc.)

I'm thinking:

- Auto renew setting.
- Currency.
- Enabled gateways.
- Number of members.
- Number of subscription levels.

What else?

pippinsplugins:
Here's what AFFWP tracks:
```php
	/**
	 * Retrieves site data (plugin versions, integrations, etc) to be sent along with the license check.
	 *
	 * @since 1.9
	 * @access public
	 *
	 * @return array
	 */
	public function get_site_data() {

		$data = array();

		$theme_data = wp_get_theme();
		$theme      = $theme_data->Name . ' ' . $theme_data->Version;

		$data['php_version']  = phpversion();
		$data['affwp_version']  = AFFILIATEWP_VERSION;
		$data['wp_version']   = get_bloginfo( 'version' );
		$data['server']       = isset( $_SERVER['SERVER_SOFTWARE'] ) ? $_SERVER['SERVER_SOFTWARE'] : '';
		$data['install_date'] = get_post_field( 'post_date', affwp_get_affiliate_area_page_id() );
		$data['multisite']    = is_multisite();
		$data['url']          = home_url();
		$data['theme']        = $theme;

		// Retrieve current plugin information
		if( ! function_exists( 'get_plugins' ) ) {
			include ABSPATH . '/wp-admin/includes/plugin.php';
		}

		$plugins        = array_keys( get_plugins() );
		$active_plugins = get_option( 'active_plugins', array() );

		foreach ( $plugins as $key => $plugin ) {
			if ( in_array( $plugin, $active_plugins ) ) {
				// Remove active plugins from list so we can show active and inactive separately
				unset( $plugins[ $key ] );
			}
		}

		$data['active_plugins']   = $active_plugins;
		$data['inactive_plugins'] = $plugins;
		$data['locale']           = get_locale();
		$data['integrations']     = affiliate_wp()->integrations->get_enabled_integrations();
		$data['affiliates']       = affiliate_wp()->affiliates->count( array( 'number' => -1 ) );
		$data['creatives']        = affiliate_wp()->creatives->count( array( 'number' => -1 ) );
[...]

PR #1322, opened 2017-07-26
Add usage tracking #1304

nosegraze:
#1304

### DIFF
--- a/includes/admin/settings/settings.php
+++ b/includes/admin/settings/settings.php
@@ -1829,6 +1829,11 @@ function rcp_get_site_tracking_data() {
 	 */
 	global $rcp_levels_db;
 
+	/**
+	 * @var RCP_Payments $rcp_payments_db
+	 */
+	global $rcp_payments_db;
+
 	$data = array();
 
 	$theme_data = wp_get_theme();
@@ -1869,6 +1874,7 @@ function rcp_get_site_tracking_data() {
 	$data['expired_members']     = rcp_get_member_count( 'expired' );
 	$data['cancelled_members']   = rcp_get_member_count( 'cancelled' );
 	$data['subscription_levels'] = $rcp_levels_db->count();
+	$data['payments']            = $rcp_payments_db->count();
 
 	return $data;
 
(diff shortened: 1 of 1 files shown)


includes/admin/settings/settings.php
```diff
@@ -1804,6 +1814,81 @@ function rcp_check_license() {
...
+	$data['inactive_plugins']    = $plugins;
+	$data['locale']              = get_locale();
+	$data['auto_renew']          = $rcp_options['auto_renew'];
+	$data['currency']            = $rcp_options['currency'];
+	$data['gateways']            = $enabled_gateways;
+	$data['active_members']      = rcp_get_member_count( 'active' );
+	$data['free_members']        = rcp_get_member_count( 'free' );
+	$data['expired_members']     = rcp_get_member_count( 'expired' );
+	$data['cancelled_members']   = rcp_get_member_count( 'cancelled' );
+	$data['subscription_levels'] = $rcp_levels_db->count();
```
mindctrl: There's not a `count()` method in `RCP_Levels`, but we can count the result of `get_levels()`.

includes/admin/settings/settings.php
```diff
@@ -1791,6 +1797,10 @@ function rcp_check_license() {
 
 		set_transient( 'rcp_license_check', $license_data->license, DAY_IN_SECONDS );
 
+		if( ! empty( $api_params['site_data'] ) ) {
+			update_option( 'rcp_last_checkin', current_time( 'timestamp' ) );
```
mindctrl: Since this is only needed once a week, we should probably pass `false` as the third param to prevent it from autoloading.

includes/admin/settings/settings.php
nosegraze: https://github.com/restrictcontentpro/restrict-content-pro/blob/release/2.9/includes/class-rcp-levels.php#L167-L175 ?

mindctrl: Ah, forgot we added that in 2.9.

PR #1345, opened 2017-08-13
Update payments-page.php

flowdee:
Adding "rcp_payments_page_table_row_actions" hook which allows adding more actions to each payment row


flowdee:
Ah sorry, didn't know. Any ETA for the v2.9?

flowdee:
Updated pull request for the latest version

mindctrl:
@flowdee This update isn't right. It's duplicating some lines. Please resubmit this as a new PR so that the only change is the single line you want added.

Also, it's probably good to change `$payment->id` to `$payment` in case someone needs the other payment details for their row action. Otherwise they'd need to duplicate the query just to get that info.

flowdee:
Alright, thanks John!

PR #1374, opened 2017-08-21
Add support for custom roles in [restrict] shortcode #1309

### LINKED ISSUE #1309: Add support for custom roles in [restrict] shortcode
We added support for custom roles in https://github.com/restrictcontentpro/restrict-content-pro/issues/158, but we did not add support for them in the [restrict] shortcode. I think we should do that.

nosegraze:
Closes #1309

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -72,27 +72,9 @@ function rcp_restrict_shortcode( $atts, $content = null ) {
 		}
 	}
 
-	if ( $atts['userlevel'] === 'admin' && ! current_user_can( 'switch_themes' ) ) {
-		$has_access = false;
-	}
-
-	if ( $atts['userlevel'] === 'editor' && ! current_user_can( 'moderate_comments' ) ) {
-		$has_access = false;
-	}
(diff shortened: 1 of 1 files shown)


includes/shortcodes.php
```diff
@@ -74,8 +74,17 @@ function rcp_restrict_shortcode( $atts, $content = null ) {
 
 	if ( $atts['userlevel'] === 'none' && ! is_user_logged_in() ) {
 		$has_access = false;
-	} elseif ( ! current_user_can( strtolower( $atts['userlevel'] ) ) ) {
-		$has_access = false;
+	} elseif( 'none' != $atts['userlevel'] ) {
```
mindctrl: I think this would be easier to read if you dropped the `elseif` and just put the `if` statement on its own line below.

nosegraze: Like this https://github.com/restrictcontentpro/restrict-content-pro/pull/1374/commits/0f6bf3a89a06cc657efb296215b9a26849fcbbdb ?

PR #1415, opened 2017-09-08
Make post-verification redirect filterable

### LINKED ISSUE #1414: Add page redirection for email verification
Each time one of my testing people activate the account, because the email verification open the edit the profile page, they believe he have to change the password again as is empty the space and is the first section they see.

Maybe to be able to choose the page that is open after the verification would create a better path for the users.

chrisvanpatten:
Adds a new filter, `rcp_verification_redirect_url`, allowing users to redirect to an arbitary URL after clicking an email verification link.

Fixes #1414.

### DIFF
--- a/includes/member-functions.php
+++ b/includes/member-functions.php
@@ -1400,7 +1400,7 @@ function rcp_confirm_email_verification() {
 		return;
 	}
 
-	wp_safe_redirect( apply_filters( 'rcp_verification_redirect_url', $redirect, $user->ID ) );
+	wp_safe_redirect( apply_filters( 'rcp_verification_redirect_url', $redirect, $member ) );
 	exit;
 
 }
--- a/includes/member-functions.php
+++ b/includes/member-functions.php
@@ -1400,7 +1400,7 @@ function rcp_confirm_email_verification() {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


chrisvanpatten:
(Not sure why the build failed on this one, but it looks like it's unrelated to the PR itself!) Any way to trigger another attempt?

pippinsplugins:
I've restarted the build!

chrisvanpatten:
Thanks @pippinsplugins! Much appreciated :)

includes/member-functions.php
```diff
@@ -1400,7 +1400,7 @@ function rcp_confirm_email_verification() {
 		return;
 	}
 
-	wp_safe_redirect( $redirect );
+	wp_safe_redirect( apply_filters( 'rcp_verification_redirect_url', $redirect, $user->ID ) );
```
nosegraze: We should include the `$member` object as the second paramter instead of `$user->ID`. The member object is a bit more useful. :)

chrisvanpatten:
Thanks @nosegraze - updated :)

nosegraze:
Thanks for the PR and for making that change @chrisvanpatten ! :)

Issue #1419, opened 2017-09-13
Customizable informations option for manual payments


brandomeniconi:
Usually manual payments are used for Bank/Wire Transfers and the user needs to be instructed where and how to make the payment.

I propose to create a field in RCP payments options where the admin can specify a text that will be printed when the user select the manual payment option.

mindctrl:
@brandomeniconi thanks for the issue and the pull request! This is something we've discussed adding for all the gateways, not just manual payments. I'm going to rename this issue to cover all the gateways and close your pull request until we can address this in a future update. Thanks!

brandomeniconi:
Beautiful! Thanks.

nosegraze:
Related: it would be nice to allow users to reorder the gateways as well.

mindctrl:
👍   both of these should probably happen after #155

PR #1420, opened 2017-09-13
Added customizable informations for manual payments.

brandomeniconi:
Allows admins to specify a text printed under the manual payment option in the registration form. See #1419


thomasfw:
Any chance of merging this, seems like a reasonable solution and there doesn't seem to be much progress since #1419/#155?

PR #1430, opened 2017-09-18
Don't try to create Braintree token for free subs #1307

### LINKED ISSUE #1307: JS error when only Braintree is activated & registering for free plan
This happens if the Braintree gateway is _technically_ selected, but then you're trying to register for a free plan. The form won't submit.

rcp-braintree

The problem doesn't occur if you have another gateway activated at the same time (like PayPal) and that other gateway is technically selected when switching to the free plan.

nosegraze:
Closes #1307

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-braintree.php
+++ b/includes/gateways/class-rcp-payment-gateway-braintree.php
@@ -749,14 +749,14 @@ class RCP_Payment_Gateway_Braintree extends RCP_Payment_Gateway {
 				}
 
 				// Get the subscription price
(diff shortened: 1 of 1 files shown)


includes/gateways/class-rcp-payment-gateway-braintree.php
```diff
@@ -748,6 +748,18 @@ public function fields() {
 					return;
 				}
 
+				// Get the subscription price
+				if( $('.rcp_level:checked').length ) {
```
mindctrl: `$` is not defined here. Will update to use `jQuery` instead.

PR #1439, opened 2017-09-21
Move rcp_form_processing hook below pending payment creation #1438

### LINKED ISSUE #1438: Move 'rcp_form_processing' hook below pending payment creation
I think this is necessary to fix https://github.com/AffiliateWP/AffiliateWP/issues/2389 That way the pending payment ID will be accessible in the hook.

nosegraze:
Closes #1438


nosegraze:
We do actually store the subscription key inside the pending payment, so it can be accessed that way.

mindctrl:
@nosegraze I think we need to fix it so existing integrations don't break like AffWP did. They shouldn't need to make any changes here. Agree?

nosegraze:
You're right, that's a good point. :)

Issue #1449, opened 2017-09-22
Allow display/edit of custom meta fields in edit-payment admin page


brandomeniconi:
There is no way to add custom fields to the [edit-payments.php](https://github.com/restrictcontentpro/restrict-content-pro/blob/bc25841396eac9f222cf3e35e2a1d1f5f809e5b5/includes/admin/payments/edit-payment.php#L127) template.

We need to add a meta field to display/edit the invoice number for an external Invoicing system that we are integrating with RCP, but actually there is no hook to display it.

PR #1450, opened 2017-09-22
Actions for custom fields in signle payment admin page. #1449

brandomeniconi:
Added following actions:

`rcp_add_payment_before`
`rcp_add_payment_after`

`rcp_edit_payment_before`
`rcp_edit_payment_after`

Allow third-party plugins to print custom meta fields in new payment and edit payment pages.

Related to #1449.

### DIFF
--- a/includes/admin/payments/edit-payment.php
+++ b/includes/admin/payments/edit-payment.php
@@ -27,7 +27,7 @@ $subscription = rcp_get_subscription_details( $payment->object_id );
 <form id="rcp-edit-payment" action="" method="post">
 	<table class="form-table">
 		<tbody>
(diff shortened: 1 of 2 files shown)


mindctrl:
@brandomeniconi @nosegraze see any reason to not ditch `$payment_id` in this hook and just use the full payment `$payment` object in its place? It's easy enough to get the payment id at `$payment->id`.

It also may be helpful to pass `$user` to it, since we already look it up and it could be useful.

something like `do_action( 'rcp_edit_payment_before', $payment, $subscription, $user );`

nosegraze:
I'm happy with your proposed changes, @mindctrl .

brandomeniconi:
I agree, PR updated as requested!

mindctrl:
@brandomeniconi thanks! Merged.

PR #1470, opened 2017-10-03
Show success message after resetting password #1364

### LINKED ISSUE #1364: Show success message after resetting password
Right now the flow is a bit like this:

1. Enter email to request password reset.
2. Get success message saying, "check your email".
3. Click link in email.
4. Get sent to form where you enter new password.
5. Get redirected to login form **with no message/notice/etc.**

That last step has this query string in the URL: `?password-reset=true` so we could easily just show a message on there saying something like, _"Your password has been successfully reset. You may now login."_ One customer noted it may not be clear to members that their password reset was successful.

nosegraze:
Closes #1364

### DIFF
--- a/templates/login.php
+++ b/templates/login.php
@@ -16,6 +16,12 @@
 global $rcp_login_form_args; ?>
 <?php if( ! is_user_logged_in() ) : ?>
 
+	<?php if ( isset( $_GET['password-reset'] ) && 'true' == $_GET['password-reset'] ) { ?>
+		<p class="rcp_success">
+			<span><?php _e( 'Your password has been successfully reset. You may now login.', 'rcp' ); ?></span>
+		</p>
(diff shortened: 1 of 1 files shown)


templates/login.php
```diff
@@ -16,6 +16,12 @@
 global $rcp_login_form_args; ?>
 <?php if( ! is_user_logged_in() ) : ?>
 
+	<?php if ( isset( $_GET['password-reset'] ) && 'true' == $_GET['password-reset'] ) { ?>
+		<p class="rcp_success">
+			<span><?php _e( 'Your password has been successfully reset. You may now login.', 'rcp' ); ?></span>
```
mindctrl: `login` should be `log in`

nosegraze: Updated.

PR #1471, opened 2017-10-03
Add image to Stripe Checkout args. #1317

### LINKED ISSUE #1317: Stripe Checkout args are inconsistent on [register_form] and [register_form_stripe]
For example, in `[register_form_stripe]` we set an image:

```php
if ( empty( $data['data-image'] ) && $image = get_site_icon_url() ) {
	$data['data-image'] = $image;
}
```

We don't do that in `[register_form]`. I think we should try to make them pretty much the same between the two options.

nosegraze:
Closes #1317


includes/gateways/class-rcp-payment-gateway-stripe-checkout.php
```diff
@@ -61,6 +61,11 @@ public function fields() {
 			'currency'          => rcp_get_currency()
 		) );
 
+		$image = get_site_icon_url();
+		if ( ! empty( $image ) ) {
+			$data['image'] = $image;
+		}
+
```
mindctrl: This will overwrite any image someone sets with the filter above it, which could potentially be a different image. I think we should go ahead and move the filter below this to avoid that possibility.

nosegraze: Updated https://github.com/restrictcontentpro/restrict-content-pro/pull/1471/commits/ff988e3fe67746d9c70380e3b204fd35e5721350

PR #1474, opened 2017-10-04
Introduce rcp_format_amount() #1384

### LINKED ISSUE #1384: Trailing zeroes removed from amount in emails
For example: `$12.50` becomes `$12.5` in the email.

nosegraze:
Closes #1384


includes/misc-functions.php
```diff
@@ -962,6 +962,24 @@ function rcp_currency_decimal_filter( $decimals = 2 ) {
 }
 
 /**
+ * Formats the payment amount for display to enforce two decimal places
+ *
```
mindctrl: This enforces the decimal places of the currency, which isn't always two.

nosegraze: Updated https://github.com/restrictcontentpro/restrict-content-pro/pull/1474/commits/9715b7fcd92eb4c5bdc710ffee73a25c6163fb1b

PR #1485, opened 2017-10-10
Remove incorrect docblock line

tnorthcutt:
`rcp_process_registration()` does not actually call `rcp_add_user_to_subscription()`

### DIFF
--- a/includes/registration-functions.php
+++ b/includes/registration-functions.php
@@ -18,8 +18,6 @@ if ( ! defined( 'ABSPATH' ) ) exit;
 /**
  * Register a new user
  *
- * @uses rcp_add_user_to_subscription()
- *
  * @access public
  * @since  1.0
  * @return void


nosegraze:
It calls it right here https://github.com/restrictcontentpro/restrict-content-pro/blob/fbb17fa188738efcdbdb746f630a0fe9a267cf0a/includes/registration-functions.php#L1284-L1286 ?

tnorthcutt:
Maybe I'm missing something, but it looks like that's `rcp_user_register_add_subscription_level()` calling `rcp_add_user_to_subscription()`. I don't see where `rcp_process_registration()` calls `rcp_add_user_to_subscription()`.

nosegraze:
Sorry my eyes jumped to the wrong function because GitHub highlighted that line in files changed. You're right. 👍

tnorthcutt:
Whew, I thought I was going crazy 😀

Issue #1486, opened 2017-10-11
Provide action after RCP install routine has run


tnorthcutt:
I'm in a situation where I need to do some stuff after RCP has been activated; specifically after `rcp_create_tables()` has run.

Currently, neither `rcp_create_tables()` nor `rcp_options_install()` have any actions to hook into. It looks like a workaround may be to use the [`activate_PLUGINNAME`](https://developer.wordpress.org/reference/functions/register_activation_hook/) hook, but that feels a bit hacky.

**Would it be reasonable to provide an e.g. `rcp_post_activation` action to hook into for cases like this?**

**Context:** my use case is that RCP is being activated on new subsites created in a multisite install, and I'm trying to populate those new sites with various things, including RCP subscription levels. Right now I'm hooking to `wpmu_new_blog`, but at that time RCP hasn't been activated yet, and so `RCP_Levels->insert()` fails b/c the DB table it's looking for doesn't exist.

mindctrl:
@tnorthcutt I was trying to think of a clever workaround for you, like hooking into `pre_update_option_rcp_db_version` just to hook into `shutdown` at priority `11` to ensure `rcp_create_tables()` has run. We only update the rcp_db_version option during installs and upgrades, so technically it would probably work fine. But I'm ok with us adding a hook inside `rcp_create_tables()` so it's easier to understand what's going on.

PR #1488, opened 2017-10-11
Add rcp_create_tables hook

tnorthcutt:
Closes https://github.com/restrictcontentpro/restrict-content-pro/issues/1486

I tried to follow what appear to be existing naming patterns for hooks in RCP; obviously it's fine if a different name is used.


tnorthcutt:
@mindctrl Added that 👍

PR #1508, opened 2017-10-17
Swap Stripe Publishable & Secret Key settings & add placeholder text #1443

### LINKED ISSUE #1443: Swap order of Publishable Key and Secret Key in settings
For consistency with Stripe's settings it would be nice to have the Publishable Key first, followed by the Secret Key.

screenshot 2017-09-22 12 29 36

screenshot 2017-09-22 12 29 56

nosegraze:
Closes #1443


includes/admin/settings/settings.php
```diff
@@ -358,38 +358,38 @@ function rcp_settings_page() {
 						</tr>
 						<tr>
 							<th>
-								<label for="rcp_settings[stripe_test_secret]"><?php _e( 'Test Secret Key', 'rcp' ); ?></label>
+								<label for="rcp_settings[stripe_test_publishable]"><?php _e( 'Test Publishable Key', 'rcp' ); ?></label>
 							</th>
 							<td>
-								<input class="regular-text" id="rcp_settings[stripe_test_secret]" style="width: 300px;" name="rcp_settings[stripe_test_secret]" value="<?php if(isset($rcp_options['stripe_test_secret'])) { echo $rcp_options['stripe_test_secret']; } ?>"/>
-								<p class="description"><?php _e('Enter your test secret key. Your API keys can be obtained from your <a href="https://dashboard.stripe.com/account/apikeys" target="_blank">Stripe account settings</a>.', 'rcp'); ?></p>
+								<input class="regular-text" id="rcp_settings[stripe_test_publishable]" style="width: 300px;" name="rcp_settings[stripe_test_publishable]" value="<?php if(isset($rcp_options['stripe_test_publishable'])) { echo $rcp_options['stripe_test_publishable']; } ?>" placeholder="<?php esc_attr_e( 'pk_test_xxxxxxxx', 'rcp' ); ?>"/>
```
mindctrl: Do all these placeholders need to be translatable? From what I've seen they're always the same format, `pk_` or `sk_`. If that's true, we could ditch the escaping and hardcode the placeholders, and avoid adding weird strings to the translation files. If they're not always that, we should probably use `esc_attr_x` and give some context here since the strings don't really communicate any context.

nosegraze: I don't _think_ they need to be translatable. We could always hard-code them in and change that later if we get requests otherwise.

PR #1520, opened 2017-10-18
Reapply custom role on membership renewal #1257

### LINKED ISSUE #1257: User role not re-applied to expired user who gets reactivated
This would only happen if the user expires and then their membership gets reactivated without them going through the registration form. Here's an example scenario a customer had:

- Have custom role applied via subscription level.
- User signs up for recurring subscription via Stripe and gets the custom role on their account.
- User's account expires due to failed payment / expired card. The custom role is removed.
- They're using Baremetrics Recover and the user updates their card details that way.
- The card gets updated in Stripe, Stripe processes the payment, and the account gets reactivated.
- But the custom role is never reapplied.

nosegraze:
Closes #1257


includes/class-rcp-member.php
```diff
@@ -443,6 +444,12 @@ public function renew( $recurring = false, $status = 'active', $expiration = ''
 		$this->set_recurring( $recurring );
 		$this->set_renewed_date();
 
+		// Add the role if the user doesn't already have it.
+		$role = ! empty( $subscription_level->role ) ? $subscription_level->role : 'subscriber';
```
mindctrl: Should we use `get_option( 'default_role', 'subscriber' )` here instead of hardcoding `subscriber`?

nosegraze: Yeah probably. I just copied it from [here](https://github.com/restrictcontentpro/restrict-content-pro/blob/master/includes/registration-functions.php#L1212) where it's hard-coded too. We can probably update both. Though honestly the `else` should never happen anyway, I don't think.

nosegraze: Updated both.

Issue #1526, opened 2017-10-19
Upgrades paid via PayPal can result in immediate subscription cancellations


mindctrl:
Have a customer seeing this sometimes. It happens due to ordering of the IPNs.

When they come in like this: `subscr_payment` and then `subscr_signup` (as opposed to the other way around), the subscription ends up getting cancelled when the `subscr_signup` comes in.

When `subscr_payment` comes in, we set the payment profile ID here: https://github.com/restrictcontentpro/restrict-content-pro/blob/2.9.5/includes/gateways/class-rcp-payment-gateway-paypal.php#L400

When `subscr_signup` comes in after `subscr_payment`, we check to see if a member just upgraded and then cancel their saved profile ID: https://github.com/restrictcontentpro/restrict-content-pro/blob/2.9.5/includes/gateways/class-rcp-payment-gateway-paypal.php#L372-L374

I think we can remove `$member->set_payment_profile_id( $posted['subscr_id'] );` from the `subscr_payment` block and it will fix this.

PR #1527, opened 2017-10-19
Don't set payment profile ID in PayPal during subscr_payment events. #1526

mindctrl:
#1526


nosegraze:
PayPal sandbox isn't working for me right now. I can try again later today.

Were you able to test this @mindctrl ?

mindctrl:
@nosegraze yes, but I never did see PayPal send the IPNs in that order. In my tests `subscr_signup` was coming in before `subscr_payment`.

nosegraze:
Well look at that... just tested again and sandbox not only worked, but the `subscr_payment` IPN even came through first. :P Worked perfectly fine.

mindctrl:
Yay!

PR #1540, opened 2017-10-26
Mark payments as refunded with webhooks #1413

### LINKED ISSUE #1413: Mark payments as refunded with webhook listener (Stripe)
The Stripe webhook listener doesn't currently process the **charge.refunded** event. It would be useful if payments refunded in Stripe were automatically marked as such in RCP - should be pretty simple to achieve with an extra `$event->type` conditional in the gateway's `process_webhooks()`.

nosegraze:
Closes  #1413

### DIFF
--- a/includes/gateways/class-rcp-payment-gateway-stripe.php
+++ b/includes/gateways/class-rcp-payment-gateway-stripe.php
@@ -684,6 +684,28 @@ class RCP_Payment_Gateway_Stripe extends RCP_Payment_Gateway {
 
 					}
 
+					// Refunded payment.
+					if ( $event->type == 'charge.refunded' ) {
+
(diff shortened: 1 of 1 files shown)


includes/gateways/class-rcp-payment-gateway-stripe.php
```diff
@@ -684,6 +684,28 @@ public function process_webhooks() {
 
 					}
 
+					// Refunded payment.
+					if ( $event->type == 'charge.refunded' ) {
+
+						rcp_log( 'Processing charge.refunded webhook.' );
```
pippinsplugins: Let's add the webhook's event ID here for easier debugging.

nosegraze: Updated!

PR #1544, opened 2017-10-31
Add more data to payments export #1142

### LINKED ISSUE #1142: Add more info to payments export
We should add:

- transaction_id
- status

And, after #1000, discount_code and gateway.

nosegraze:
Closes #1142


nosegraze:
Updated to add them, though I named `object_id` as `Subscription ID` rather than `Object ID` since we only support subscriptions right now. I reckon we can change the name of that if/when we add other types.

PR #1576, opened 2017-11-15
Don't require a paid account with level restrictions #1569

### LINKED ISSUE #1569: Free user can't view post restricted to "specific levels" if one is paid
Set up post restrictions for `Members of specific subscription levels` and select at least two levels: one free one and one paid. When a user is on the paid plan they can see the content, but when a user is on the free plan they cannot.

See this block: https://github.com/restrictcontentpro/restrict-content-pro/blob/master/includes/class-rcp-member.php#L1313-L1318

Basically if one of the levels is paid, then it suddenly becomes a requirement that the user be on a paid level to view the content.

Is there a reason we need that block at all?

nosegraze:
Closes #1569

### DIFF
--- a/includes/admin/metabox.php
+++ b/includes/admin/metabox.php
@@ -149,13 +149,15 @@ function rcp_save_meta_data( $post_id ) {
 
 				case 'specific' :
 
+					$is_paid = true;
+
 					$levels = array_map( 'absint', $_POST[ 'rcp_subscription_level' ] );
 
 					foreach( $levels as $level ) {
 
 						$price = rcp_get_subscription_price( $level );
-						if( ! empty( $price ) ) {
-							$is_paid = true;
+						if( empty( $price ) ) {
+							$is_paid = false;
 							break;
 						}
 
@@ -177,16 +179,6 @@ function rcp_save_meta_data( $post_id ) {
 
 			update_post_meta( $post_id, 'rcp_access_level', absint( $_POST[ 'rcp_access_level' ] ) );
 
-			$levels = rcp_get_subscription_levels();
-			foreach( $levels as $level ) {
-
-				if( ! empty( $level->price ) ) {
-					$is_paid = true;
-					break;
-				}
-
-			}
-
 			// Remove unneeded fields
 			delete_post_meta( $post_id, 'rcp_subscription_level' );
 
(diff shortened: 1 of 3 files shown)


mindctrl:
That test doesn't appear to be valid really, since the user has the level required to access content. There's no use case where the user has a paid level and ends up with a free status.

nosegraze:
👍

mindctrl:
I think `test_cannot_access_active_content_as_free_member()` covers the intended use for `test_cannot_access_active_content_as_free_member_with_subscription_level()` (see original test, not the updated one here). The latter doesn't seem like a good test, since we don't set a `free` status on members that have paid levels. I don't recall why we originally had the paid check loop and this test. @pippinsplugins do you?

nosegraze:
I think your'e right, but maybe we also need to look at when we're actually saving `_is_paid` as `true` in the meta:

```php
case 'specific' :

	$levels = array_map( 'absint', $_POST[ 'rcp_subscription_level' ] );

	foreach( $levels as $level ) {

		$price = rcp_get_subscription_price( $level );
		if( ! empty( $price ) ) {
			$is_paid = true;
			break;
		}

	}

	update_post_meta( $post_id, 'rcp_subscription_level', $levels );

	break;
```

So once again, if _any_ of the specific levels are paid, then we set `_is_paid` to true. Presumably what we actually want to do is default `$is_paid` to `true` in this block, then check if any of the chosen subscription levels are free, and if so, change `$is_paid` to `false` ?

So if _all_ of the chosen levels are paid, then `_is_paid` should be `true`.
If _any_ of them are free, then `_is_paid` should be `false`.

mindctrl:
@nosegraze I think you're right. There's also similar logic in the other cases, where it sets `_is_paid` if any paid sub levels exist at all, even if the restriction settings are tied to roles or access levels.

I don't remember all the history about why that was the case, but I'd love to rip it all out if we can. Updating the PR here in a few.

mindctrl:
Just pushed the update that I meant to send last night. @nosegraze do you foresee any issues with that?

nosegraze:
This looks good to me.

The only downside is that if access level or role is used, `_is_paid` will just always be `false`. However, given that access levels and roles are far more complex and can span over multiple different levels, I think I'm fine with that.

mindctrl:
@nosegraze can you give this one more look and test? If all goes well, let's merge it.

nosegraze:
Yes I'll take a closer look today. I also just sent the changes to a customer who ran into this issue. I'm waiting to hear back to see if it fixed the problem for him.

nosegraze:
Tested and confirmed the following:

- [x] When selecting two levels - one free and one paid - the `_is_paid` meta is not set. User on free level can access. User on paid level can access.
- [x] When selecting one level - paid level - the `_is_paid` meta is set. User on paid level can access.
- [x] When selecting one level - free level - the `_is_paid` meta is not set. User on free level can access.
- [x] When selecting three levels - all different paid ones - the `_is_paid` meta is set. User on one of the paid levels can access.
- [x] When selecting two levels - different free ones - the `_is_paid` meta is not set. User on one of the free levels can access.

Issue #814, opened 2016-12-08
Replace "prevent account sharing" with a maximum number of simultaneous connections 


geoffreybr:
Our typical user uses our website on his personal computer, on his smartphone and on his work computer. The current option let us choose between allowing a single connection (so the user has to connect every time he switches between his devices, which is painful) and allowing an unlimited number of connections (so the user can share his account).

It could be nice to replace the checkbox with a number input to let the admin choose the number of simultaneous connections. I think this number simply need to be used in line 431 of *includes/misc-functions.php*. A WP filter on it could be great too.
Moreover, it could be nice to customize it with a parameter in the subscription level in order to create one-connection subscription level, and a more expensive ten-connection subscription level for example. I already did this in my own plugin to create a more flexible subscription level allowing account sharing, using the `rcp_no_account_sharing` filter.

*At the moment, I do not have enough time to make a pull request and test it properly. Maybe later, but I put it here if you are faster than me :grinning:...*

pippinsplugins:
I like this idea.

geoffreybr:
@mindctrl Do you think this could be integrated before 2.8 ? I can check if there are conflicts with 2.7 if that's helpful.

mindctrl:
@geoffreybr we're aiming to release 2.8 by end of this month - we're moving to shorter, smaller release cycles.

geoffreybr:
I just made a new PR without the parameter in the subscription level. Should it be possible to review it and include it in the next version?
I really need it for 3 of my clients (so 3 of yours too :smiley:). Thanks!

PR #1578, opened 2017-11-17
#814 - Replace 'Prevent account sharing' with a 'Maximum number of simultaneous connections'

geoffreybr:
A lighter version of #890 with the replacement of the bool by a number only.

This may need further testing but it works for 0 (unlimited), 1, 2, 3 simultaneous connections.
This should be compatible with previous version as the previous parameter was '1' for account sharing prevented and 'not set' for account sharing allowed.

Refers to issue #814


includes/misc-functions.php
```diff
@@ -326,15 +326,27 @@ function rcp_get_current_url() {
...
+ * Returns the "Maximum number of simultaneous connections per member".
+ *
+ * @access private
+ * @since  2.7
+ * @return int
+*/
+function rcp_no_account_sharing_number() {
 	global $rcp_options;
-	return (bool) apply_filters( 'rcp_no_account_sharing', isset( $rcp_options['no_login_sharing'] ) );
+	return apply_filters( 'rcp_no_account_sharing', $rcp_options['no_login_sharing'] );
```
nosegraze: If someone previously had the option unchecked, this would throw a PHP notice:

```
Undefined index: no_login_sharing
```

You need to check if it's set first.

geoffreybr: Fixed in e7aca86ba785681353981c9f04c85c234cb8a528

Issue #192, opened 2015-06-26
Add batch export API


pippinsplugins:
Just like we added in EDD, RCP should have a batch export API to allow it to handle large amounts of data.

pippinsplugins:
We have completed the batch processing API for AffiliateWP now too. It's more robust than EDD's so we should use it as a model.

https://github.com/AffiliateWP/AffiliateWP/issues/1707

pippinsplugins:
Really awesome library from Reaktiv https://github.com/reaktivstudios/locomotive

mindctrl:
At some point we need to support wp-cli batching.

PR #1612, opened 2017-12-20
Introduce batch processing. #192

mindctrl:
WIP: #192


mindctrl:
Going to update this to support running multiple batches in succession, so we can offer a better user experience and not require people run them manually. This will be nice for 3.0.

PR #1614, opened 2017-12-21
Remove usermeta and custom capabilities on uninstall. #1229

### LINKED ISSUE #1229: Remove user metadata during uninstall
After deleting RCP with the setting to remove RCP data checked I can still find a lot of rcp_ prefixed metadata in

- User roles
- User meta

nosegraze:
Closes #1229

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -34,6 +34,13 @@ if( isset( $rcp_options['remove_data_on_uninstall'] ) ) {
 	// Delete all term meta.
 	$wpdb->query( "DELETE FROM $wpdb->termmeta WHERE meta_key = 'rcp_restricted_meta'" );
 
+	// Delete all user meta.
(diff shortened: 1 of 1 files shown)


uninstall.php
```diff
@@ -34,6 +34,13 @@
 	// Delete all term meta.
 	$wpdb->query( "DELETE FROM $wpdb->termmeta WHERE meta_key = 'rcp_restricted_meta'" );
 
+	// Delete all user meta.
+	$wpdb->query( "DELETE FROM $wpdb->usermeta WHERE meta_key LIKE 'rcp\_%'" );
```
mindctrl: We also need to delete `_rcp_%` to get all the ones related to emails, new/old subscription, etc.

nosegraze: Good catch. Updated https://github.com/restrictcontentpro/restrict-content-pro/pull/1614/commits/5c29137ffbbc29e727cdc767d1c2bfb87c5f7238

PR #1634, opened 2018-01-22
Exit function if RCP settings ID doesn't exist on page. #1633

### LINKED ISSUE #1633: JavaScript conflict with other plugins that use hashes in URLs
Report: https://secure.helpscout.net/conversation/509668960/74514?folderId=841781

This is reproducible with the Envira Gallery plugin, which uses a tabbed interface on their custom post type and changes the `#tab-id` in the URL when the tab changes. This change triggers the JavaScript in RCP that looks for hash changes in the URL for RCP's tab interface. But it causes a JavaScript error when `"#rcp-settings-wrap #tab_container #" + url` doesn't exist.

nosegraze:
Closes #1633


includes/js/admin-scripts.js
```diff
@@ -94,6 +94,10 @@ jQuery(document).ready(function($) {
 	//when the history state changes, gets the url from the hash and display
 	$(window).bind( 'hashchange', function(e) {
 
+		if ( $('#rcp-settings-wrap').length == 0) {
+			return;
+		}
+
```
mindctrl: We can avoid another DOM lookup here by using `window.adminpage` like this:

```
		if( ! window.adminpage || ( window.adminpage && 'restrict_page_rcp-settings' !== window.adminpage ) ) {
			return;
		}
```

nosegraze: Nice one! Updated!

Issue #1679, opened 2018-02-16
Search functionality makes it hard to find members


tnorthcutt:
If I have a member who has `wp_usermeta` values of `Judy` & `Smith` for the keys `first_name` and `last_name`, searching for `judy` or `smith` does not find this member when using the search form on `/wp-admin/admin.php?page=rcp-members`.

But, using the search form on `/wp-admin/users.php` does find the user account for this member.

Related, but a little different: when searching for a member who is expired, it's not immediately obvious that the user has been found, unless the person searching knows to scan the 5 status groups at the top of the page:

![rcpexpiredsearch](https://user-images.githubusercontent.com/796639/36330038-bff6c6f8-132d-11e8-9412-db1fee523137.jpg)

mindctrl:
When we fix the name searching, I'd like to ditch the status filters at the top on search results views. I've seen people confused by this numerous times, and it's annoying to do an extra click to get to the results.

tnorthcutt:
@mindctrl what about defaulting the search results to show all statuses, but still allow filtering? This would depend on https://github.com/restrictcontentpro/restrict-content-pro/issues/1562

mindctrl:
I'm ok with that, just trying to think of a use case where searching for something results in a lot of results that you want to filter by status. Seems like an uncommon use case, but I definitely could be wrong.

tnorthcutt:
Yeah, fair point.

chrisvanpatten:
I think the problem increases as your community grows, and the less narrow your search is.

Another factor to consider: with multiple subscriptions on the roadmap, how will that affect filters? You might have users with subscriptions in multiple statuses.

mindctrl:
That's a good point. We've discussed this internally, but the concept of status being tied to a member is probably going to have to go away. It doesn't really make sense with multiple subscriptions, and in some cases it would be tricky or impossible to "choose" a status to return.

A member is always a member... just a way to designate someone has some kind of membership. The membership status is the thing that matters.

chrisvanpatten:
Yeah, it makes sense that in the future there might be separate "Members" and "Subscriptions" screens, similar to how Stripe handles it.

tnorthcutt:
[...]

PR #1702, opened 2018-03-03
Alter member search to be more inclusive

tnorthcutt:
This change mimics WP core's method of prepending and appending `*` to user searches, so that when the `WP_User_Query` is prepared, the `LIKE` clauses will be surrounded by `%`. This results in finding e.g. a member with the email address `john123@gmail.com` when searching for `john`; previously this member would not be found.

Closes https://github.com/restrictcontentpro/restrict-content-pro/issues/1679


includes/admin/members/members-page.php
```diff
@@ -29,6 +29,10 @@ function rcp_members_page() {
 			$order           = ! empty( $_GET['order']  )  ? urldecode( $_GET['order']  ) : 'DESC';
 			$search          = ! empty( $_GET['s'] )       ? urldecode( $_GET['s'] )      : '';
 
+			if ( '' !== $search ) {
+			    $search = '*' . $search . '*';
+            }
```
pippinsplugins: Looks like a minor indention hiccup here :)

tnorthcutt: Good catch! Good old tabs/spaces got me.

includes/admin/members/members-page.php
```diff
@@ -102,7 +106,7 @@ function rcp_members_page() {
 			</ul>
 			<form id="rcp-member-search" method="get" action="<?php menu_page_url( 'rcp-members' ); ?>">
 				<label class="screen-reader-text" for="rcp-member-search-input"><?php _e( 'Search Members', 'rcp' ); ?></label>
-				<input type="search" id="rcp-member-search-input" name="s" value="<?php echo esc_attr( $search ); ?>"/>
+                <input type="search" id="rcp-member-search-input" name="s" value="<?php echo isset($_REQUEST['s']) ? esc_attr( wp_unslash( $_REQUEST['s'] ) ) : ''; ?>"/>
```
nosegraze: This is indented with spaces but it should be tabs.

tnorthcutt: Sorry about that. Fixed up now.

tnorthcutt:
Thanks @mindctrl. I'm not familiar with searching by payment profile. Can you give me a little more info/an example to reproduce with? I'd be happy to fix it up.

mindctrl:
@tnorthcutt you can search members by their payment profile ID, which you can see on the member edit screen. In the case of Stripe, it's their customer number.

image

You can search for `payment_profile:xxxxxxxxx` to find the member with that profile ID.

tnorthcutt:
@mindctrl Ah, thanks! This was also breaking searches for `first_name:` and `last_name:`. I moved to altering the query to add asterisks after we check for those special cases.

Issue #1562, opened 2017-11-09
Add default "All" filter in Restrict > Members


nosegraze:
This would show all members regardless of status. This would have two benefits:

1. It would allow us to publish the total number of members in the database.
2. The default screen admins land on would show all statuses. We've had several people do a test signup, click onto that page, see "no members" and get confused, when in reality their test signup was tucked away in the "free" filter (or sometimes "pending").

nosegraze:
It might be easier to hold off on this until after https://github.com/restrictcontentpro/restrict-content-pro/issues/103 if we're still going to put that in `3.0`?

mindctrl:
I think this concept of "member status" is going away in 3.0. [See this comment](https://github.com/restrictcontentpro/restrict-content-pro/issues/1679#issuecomment-367764670).

Think about this scenario. A member has both an active and a pending subscription. What should we show as counts? And should we show the same member in both filtered views?

Related: #1679

PR #1703, opened 2018-03-03
Add default "All" filter in Restrict > Members

tnorthcutt:
This PR adds a default "all" filter in Restrict > Members.

I suggest we also increase the visual prominence of the six status filters on that page and reposition them a bit for more clarity; I'll open a separate PR for that once this is merged to avoid merge conflicts.

Note: currently, if a pending member exists without an `rcp_subscription_level` meta value, the member counts won't add up correctly; "All" will have the correct number, but "Pending" will be short, so adding up the counts of the five statuses will be less than what is shown next to "All". https://github.com/restrictcontentpro/restrict-content-pro/pull/1695 will fix this.

Closes https://github.com/restrictcontentpro/restrict-content-pro/issues/1562


tnorthcutt:
@mindctrl fair enough. Definitely a discussion worth having. FWIW I don't really see the harm in adding this in  now, especially if moving away from statuses is going to take some time (which I suspect it will), since this solves a current problem:

>The default screen admins land on would show all statuses. We've had several people do a test signup, click onto that page, see "no members" and get confused, when in reality their test signup was tucked away in the "free" filter (or sometimes "pending").

But I won't be upset if this gets paused 🙂

tnorthcutt:
Closing this in favor of https://github.com/restrictcontentpro/restrict-content-pro/pull/1777

Issue #1686, opened 2018-02-22
Post type archive redirect doesn't work with Drip Content


nosegraze:
In _redirects.php_ we have this statement for redirecting away from a post type archive:

```php
elseif( is_post_type_archive() && $wp_query->have_posts() && rcp_is_restricted_post_type( get_post_type() ) && ! $member->can_access( get_the_ID() ) ) 
```

This is meant to check if the archive being viewed is restricted, and the assumption is that if the user can't view the first post they wouldn't be able to view any.

Normally this would work fine but it doesn't work with Drip Content, where a user may have permission to view the post type but not a specific post due to the scheduling.

tnorthcutt:
One solution to this might be to short-circuit that logic if the Drip Content add-on is active.

Another option would be to do similar to what is currently done for per-post restrictions if the entire post type is restricted:

> This entire post type is restricted. Visit the post type's restriction page to change the settings.

So, when setting Drip access for a post, we could display a similar message if the post type is restricted. But, perhaps that would be counter to how people tend to use the Drip Content add-on? I'm not sure.

mindctrl:
At the moment I can't think of a way to safely do it with any `can_access()` checks without checking every post shown in the archive, which is crazy. This could also happen with any customizations, and probably Restrict Past Content, Restriction Timelock, and Restriction Timeouts. depending on the sort order.

I'm thinking we'll need to do a `rcp_get_post_type_restrictions( get_post_type() )` call and check the return values against the member's data.

nosegraze:
I think you're right @mindctrl . It's kind of annoying to have to do that, but it sounds like the best way.

PR #1704, opened 2018-03-03
Exclude restricted post types from queries

tnorthcutt:
This gets all restricted post types, builds an array of post IDs in those post types, and adds that to the `posts__not_in` query var.

Just like `rcp_get_restricted_post_ids()` and `rcp_get_post_ids_assigned_to_restricted_terms()`, the result is stored in a transient for caching.

A different approach to this could be to instead refactor `rcp_get_restricted_post_ids()` to include the post IDs belonging to restricted post types; I'm happy to make that change if we want the logic there instead of in `rcp_hide_premium_posts()`;

But, I'm not sure if that might have unintended side effects. The only other place I see `rcp_get_restricted_post_ids()` used is in `RCP_EDD::edd_downloads_query()`. On that note, we may need to make this change in that method as well (if we don't combine the logic into `rcp_get_restricted_post_ids()`), since in theory that method could miss the entire `download` post type being restricted.


includes/query-filters.php
```diff
@@ -25,13 +25,41 @@ function rcp_hide_premium_posts( $query ) {
...
+						$redirect = home_url();
+					}
+					wp_redirect( $redirect );
+					exit;
+				}
+				$query->set( 'post_type', $new_types );
+			}
+			// If one post type is queried, and it is a restricted post type, ignore query and redirect to page for redirect_from_premium
+			// Or, if no post type is specified, assume it is 'post' and redirect as needed
+			else if ( in_array( $query->get( 'post_type' ), $restricted_post_types ) || ( empty( $query->get( 'post_type' ) ) && in_array( 'post', $restricted_post_types ) ) ) {
```
nosegraze: This isn't compatible with PHP 5.3. You'll need to put the result of `$query->get( 'post_type' )` into a variable before using it.

nosegraze:
The general logic here for getting/unsetting post types is good but I'm wondering if for organizational purposes we should put the redirect code in `redirects.php` instead. So instead of actually redirecting in `query-filters.php` we could just do a `return` knowing it'll get picked up in `redirects.php` and move the redirect portion of the logic to that file.

Another thing to think about is whether we want to actually check to see if the current user has permission instead of only applying the logic to expired users or guests. See here: https://github.com/restrictcontentpro/restrict-content-pro/issues/1686#issuecomment-368758229 So instead of only removing restricted post type for users who have no membership at all, we actually compare the user's permissions against the post type restrictions. This means the logic would still work if a user has a membership for level "Silver" and a post type is restricted to level "Gold" only. All the Gold posts would then be removed for the user on Silver. Does that make sense?

includes/query-filters.php
tnorthcutt: Ah right, thanks.

tnorthcutt:
@nosegraze yep, that all makes sense.

Perhaps it would be good to split this out into multiple functions for the various restriction checking? We could have a function each for checking:
1. Individual posts with restrictions set
2. Taxonomy terms with restrictions set
3. Post types with restrictions set

Each of those methods could be hooked to `pre_get_posts` and modify the query as needed. Or, we could hook a wrapper function (perhaps the current `rcp_hide_premium_posts()`, and call the others from that.

What do you think? I'm hoping that would 1) make it easier to reason about and 2) make it easier to maintain and modify over time.

tnorthcutt:
@nosegraze any thoughts on my last comment above?

nosegraze:
I think I'm okay with that.

tnorthcutt:
On reflection, I think this needs to wait for https://github.com/restrictcontentpro/restrict-content-pro/issues/1736 to be complete first.

That is, if we want to do (which I think we should):

> Another thing to think about is whether we want to actually check to see if the current user has permission instead of only applying the logic to expired users or guests. See here: #1686 (comment) So instead of only removing restricted post type for users who have no membership at all, we actually compare the user's permissions against the post type restrictions. This means the logic would still work if a user has a membership for level "Silver" and a post type is restricted to level "Gold" only. All the Gold posts would then be removed for the user on Silver. Does that make sense?

PR #1705, opened 2018-03-03
Show error messages on Stripe checkout registration

### LINKED ISSUE #1669: [register_form_stripe] - Bad UX if user is not logged in but enters an email for an account that already exists
Visit the `[register_form_stripe]` while logged out, but then sign up with an email address that already exists. The page simply refreshes. No error messages are shown and no subscription is added.

tnorthcutt:
This shows the error messages, and loads our forms css to properly style the error messages.

Closes #1669

### DIFF
--- a/includes/registration-functions.php
+++ b/includes/registration-functions.php
@@ -434,7 +434,7 @@ function rcp_validate_user_data() {
 		if( email_exists( $user['email'] ) ) {
 			//Email address already registered
 			rcp_errors()->add( 'email_used', sprintf(
-				__( 'This email has already been registered. If this is your account you can <a href="%s">log in</a> and try again.', 'rcp' ),
+				__( 'This email address is already in use. If this is your email address, please <a href="%s">log in</a> and try again.', 'rcp' ),
 				esc_url( rcp_get_login_url() )
 			), 'register' );
 		}
--- a/includes/registration-functions.php
+++ b/includes/registration-functions.php
@@ -433,10 +433,9 @@ function rcp_validate_user_data() {
 		}
 		if( email_exists( $user['email'] ) ) {
 			//Email address already registered
-			rcp_errors()->add( 'email_used', sprintf( '%s <a href="' . esc_url( rcp_get_login_url() ) . '">%s</a> %s.',
-				__( 'This email has already been registered. If this is your account you can', 'rcp' ),
-				__( 'log in', 'rcp' ),
-				__( 'and try again', 'rcp' )
+			rcp_errors()->add( 'email_used', sprintf(
+				__( 'This email has already been registered. If this is your account you can <a href="%s">log in</a> and try again.', 'rcp' ),
+				esc_url( rcp_get_login_url() )
 			), 'register' );
 		}
 		if( empty( $user['password'] ) ) {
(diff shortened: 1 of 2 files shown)


nosegraze:
This is working well. 👍  While we're here, I wonder if we should improve the `Email already registered` error message. Maybe more like: `This email has already been registered. If this is your account you can <log in> and try again.` (with a link to the login page)

tnorthcutt:
@nosegraze great idea. Added that.

Note: the `rcp.pot` file will need to be updated. I'm not sure what process you use to generate that; if you'd like to fill me in I'd be happy to handle it (if appropriate at this point in the build process).

includes/registration-functions.php
```diff
@@ -433,7 +433,11 @@ function rcp_validate_user_data() {
 		}
 		if( email_exists( $user['email'] ) ) {
 			//Email address already registered
-			rcp_errors()->add( 'email_used', __( 'Email already registered', 'rcp' ), 'register' );
+			rcp_errors()->add( 'email_used', sprintf( '%s <a href="' . rcp_get_login_url() . '">%s</a> %s.',
```
nosegraze: `rcp_get_login_url()` needs to be escaped: `esc_url( rcp_get_login_url() )`

includes/registration-functions.php
```diff
@@ -433,7 +433,11 @@ function rcp_validate_user_data() {
...
+			                                          __( 'This email has already been registered. If this is your account you can', 'rcp' ),
```
nosegraze: Should be using tabs here (and on next two lines).

mindctrl:
@tnorthcutt we'll do the .pot file right before release.

includes/registration-functions.php
tnorthcutt: 👍

includes/registration-functions.php
tnorthcutt: 👍

tnorthcutt:
> I understand you're probably formatting it this way to avoid using the link HTML in the actual translation, but I wonder if breaking up the sentence into multiple lines like this actually makes it trickier to translate because the sentence flow is all split up?

I tweaked this so the only placeholder is for the dynamically generated login url. How's that look?

mindctrl:
Looks better 👍 . What you think about changing the text slightly to a more active voice? Instead of `This email has already been registered.` I was thinking `This email is already in use.` or something like that. I'm fine with either, just a thought I had as I read it.

tnorthcutt:
@mindctrl yeah, I like that. Technically the email address isn't what's registered (the account is), so your wording makes more sense overall. I'll tweak it.

How about:
`This email address is already in use. If this is your email address, you can log in.`

mindctrl:
Since logging in is a requirement in this case, I think we need to adjust that a little on the end. Something like this? `This email address is already in use. If this is your email address, please log in and try again.`

tnorthcutt:
@nosegraze Do those changes look ok? Not sure what the proper way is (if any) to mark a review as resolved. Should I click "Dismiss review" here?

screen shot on 2018-03-23 at 14 20 17

nosegraze:
@tnorthcutt Sorry about that - thanks for the nudge! I'm happy with this now. 👍 

(And sorry but I have no idea with regards to your review question lol. I just barely get by with the GitHub review features. :P )

tnorthcutt:
@nosegraze glad I'm not the only one 😆

PR #1710, opened 2018-03-06
Move to a global restricted content message

### LINKED ISSUE #1701: Move away from differentiating between free and paid content messages
In #914 it was suggested that we move away from differentiating between free and paid content messages. Because content can be restricted in several different ways, and can be restricted to both free and paid levels, having separate messages for free and paid content can be confusing and inaccurate.

A major factor to consider will be backwards compatibility. How do we make this change, without negatively affecting existing sites that use both free and paid messages?

Options:
1. Wait until a major version bump (3.0, 4.0, etc.), remove the dual messaging at that time, and attempt to notify site owners/operators of the change.
2. Make the change opt-in for existing sites, and on by default for new installs. IMO, this would just kick the can down the road, so to speak.
3. ???

I'm sure there are other options I'm missing right now.

Also worth discussing: are there other places, besides messaging, where we can or should move away from differentiating between free and paid?

tnorthcutt:
Adds new `global_message` setting, and uses this message if it is set.

Also prompts admins to set the new message with a dismissible notice.

If old free or paid messages are set and a new global message is not set, will fall back to using the old messages.

This introduces `rcp_get_restricted_content_message()`, which has an optional `paid` parameter. In the future, once we're past the grace period discussed in #1701, we can simply alter this function to return the global message or a default, rather than checking for the free or paid messages.

Closes #1701


includes/admin/admin-notices.php
```diff
@@ -49,6 +49,14 @@ function rcp_admin_notices() {
 			echo '<p><a href="' . wp_nonce_url( add_query_arg( array( 'rcp_notice' => 'missing_license' ) ), 'rcp_dismiss_notice', 'rcp_dismiss_notice_nonce' ) . '">' . _x( 'Dismiss Notice', 'License', 'rcp' ) . '</a></p>';
 			echo '</div>';
 		}
+
+		if ( empty( $rcp_options['global_message'] ) && ! get_user_meta( get_current_user_id(), '_rcp_content_restriction_message_missing_dismissed', true ) ) {
```
nosegraze: I think I'd prefer naming this field something like `restriction_message` over `global_message`.

includes/admin/settings/settings.php
```diff
@@ -252,28 +252,17 @@ function rcp_settings_page() {
 								<span><?php _e( 'Check this to have the auto renew checkbox enabled by default during registration. Customers will be able to change this.', 'rcp' ); ?></span>
 							</td>
 						</tr>
-						<tr valign="top">
-							<th>
-								<label for="rcp_settings[free_message]"><?php _e( 'Free Content Message', 'rcp' ); ?></label>
-							</th>
-							<td>
+                        <tr valign="top">
```
nosegraze: Spaces are being used here to indent. Needs to be tabs.

includes/misc-functions.php
```diff
@@ -1192,3 +1192,22 @@ function rcp_log( $message = '', $force = false ) {
 	$logs = new RCP_Logging();
 	$logs->log( $message );
 }
+
+/**
+ * Get restricted content message
+ *
```
nosegraze: Can you please include the `$paid` parameter in the PHPDocs? Also I'm wondering if we should move this function to `content-filters.php`, as that's where all the content-related messages are.

includes/admin/settings/settings.php
```diff
@@ -252,28 +252,17 @@ function rcp_settings_page() {
...
-							<td>
-								<?php
-								$paid_message = isset( $rcp_options['paid_message'] ) ? $rcp_options['paid_message'] : '';
-								wp_editor( $paid_message, 'rcp_settings_paid_message', array( 'textarea_name' => 'rcp_settings[paid_message]', 'teeny' => true ) ); ?>
-								<p class="description"><?php _e( 'This is the message shown to users that do not have privilege to view premium content.', 'rcp' ); ?></p>
-							</td>
-						</tr>
+								$global_message = isset( $rcp_options['global_message'] ) ? $rcp_options['global_message'] : '';
+								wp_editor( $global_message, 'rcp_settings_global_message', array( 'textarea_name' => 'rcp_settings[global_message]', 'teeny' => true ) ); ?>
+                                <p class="description"><?php _e( 'This is the message shown to users that do not have privilege to view content.', 'rcp' ); ?></p>
```
nosegraze: I know this is what was being used before but I feel like I'm only just now realizing I really don't like the wording `users that do not have privilege`. I'd prefer `users that do not have permission`.

includes/admin/admin-notices.php
tnorthcutt: Cool, will do. I figured global wasn't quite right 😁

includes/misc-functions.php
tnorthcutt: I debated putting it there, but decided to keep that file consistent to only including actual filters (with the exception of `rcp_format_teaser()` which is there). I don't love putting things in a "misc" file, but figured that's better than putting something that's not a filter in a file called filters ¯\_(ツ)_/¯

includes/admin/settings/settings.php
tnorthcutt: Totally agreed. I'm also going to alter it to `who` instead of `that.

tnorthcutt:
I think this is ready for merge. @pippinsplugins this is a somewhat significant admin UI change, if you want to take a peek.

includes/misc-functions.php
```diff
@@ -1192,3 +1192,24 @@ function rcp_log( $message = '', $force = false ) {
...
+	$message = ! empty( $rcp_options['free_message'] ) ? $rcp_options['free_message'] : false; // message shown for free content
+
+	if ( rcp_is_paid_content( $post->ID ) || in_array( $post->ID, rcp_get_post_ids_assigned_to_restricted_terms() ) || $paid ) {
+		$message = ! empty( $rcp_options['paid_message'] ) ? $rcp_options['paid_message'] : false; // message shown for premium content
+	}
+
+	$message = ! empty( $rcp_options['restriction_message'] ) ? $rcp_options['restriction_message'] : $message; // new global message
+	$message = ! empty( $message ) ? $message : __( 'This content is restricted to subscribers', 'rcp' );
+
+	return $message;
```
pippinsplugins: I think this needs to have a filter so that it's possible for developers to change this message dynamically.

tnorthcutt: @pippinsplugins yep, good call. Updating.

mindctrl:
I think we need to change the logic in this function. Since we're moving to a single message, regardless of free/paid status, if this new option is set we should return it early instead of running all the other query heavy logic to then finally return the option if it's set.

Something like this (untested):

```
function rcp_get_restricted_content_message( $paid = false ) {
	global $post, $rcp_options;

	if( ! empty( $rcp_options['restriction_message'] ) ) {
		return apply_filters( 'rcp_restricted_content_message', $rcp_options['restriction_message'] );
	}

	$message = __( 'This content is restricted to subscribers', 'rcp' );

	if( ! empty( $rcp_options['free_message'] ) ) {
		$message = $rcp_options['free_message'];
	}

	if ( ! empty( $rcp_options['paid_message'] ) && ( rcp_is_paid_content( $post->ID ) || in_array( $post->ID, rcp_get_post_ids_assigned_to_restricted_terms() ) || $paid ) ) {
		$message = $rcp_options['paid_message'];
	}

	return apply_filters( 'rcp_restricted_content_message', $message );
}
```

mindctrl:
Also in the EDD integration, it runs a paid check but then this new function runs it again. Moving the `$paid` check in front of those functions should bypass the need for them to run again.

tnorthcutt:
@mindctrl wouldn't it make more sense to keep the paid check in `rcp_get_restricted_content_message()`, so we have a single place we do that, instead of multiple others?

In the past, every place we needed a message had to do that check, since fetching the right message wasn't abstracted; now that we have a function for this, we can probably drop that. For instance in `class-rcp-easy-digital-downloads.php`:

```
if ( rcp_is_paid_content( $post_id ) ) {
    $excerpt = rcp_get_restricted_content_message( true );
} else {
    $excerpt = rcp_get_restricted_content_message();
}
```
Can probably become:
```
$excerpt = rcp_get_restricted_content_message();
```
The one place I'm not sure about is in `includes/shortcodes.php`:
```
} elseif ( $atts['paid'] ) {
	$teaser = rcp_get_restricted_content_message( true );
} else {
	$teaser = rcp_get_restricted_content_message();
}
```

What do you think?

mindctrl:
@tnorthcutt I wasn't suggesting we remove the check. It's in the example code I posted above. I agree that we should remove the paid checks that happen outside the new function, since the function does the check.

tnorthcutt:
@mindctrl roger that; I misunderstood your wording. My bad 😃 

I'll work on removing the unneeded checks.

tnorthcutt:
@mindctrl what are your thoughts on removing the check in our shortcode handler?

tnorthcutt:
👋 @mindctrl just want to make sure this didn't slip through the cracks:
> what are your thoughts on removing the check in our shortcode handler?

Issue #1003, opened 2017-02-20
Redesign the Edit Member page


nosegraze:
It could use a bit of polish.

Particularly when #774 happens.

nosegraze:
Inspiration from EDD / EDD Recurring, with a few tweaks.

![rcp-new-edit-member](https://cloud.githubusercontent.com/assets/6324272/23220277/61150f66-f919-11e6-8a2b-3262a41ddeb9.png)

![rcp-new-subscription-details](https://cloud.githubusercontent.com/assets/6324272/23220329/8a9f8bd6-f919-11e6-899c-7ffa4830f714.png)

pippinsplugins:
Oh yeah, love this!

nosegraze:
I've started playing around with this, so if there's anything you want to see added or changed from the above mockup, let me know.

pippinsplugins:
No changes at this time!

chrisvanpatten:
I'd love to see Payments and Subscription info have their own tabs, and it seems like Renewal Payments and regular Payments have some duplication. Overall though I love the direction. Reminds me a bit of JJJ's work with https://wordpress.org/plugins/wp-user-profiles/

nosegraze:
WIP PR: #1539

Edit Member `Profile` page is just about done. Notes don't work yet and `View Details` on the subscription doesn't work yet.

nosegraze:
I _love_ the notices that pop up in EDD Recurring when you go to edit certain items of the subscription. For example, when you click "Edit" on the expiration date, this notice pops up:

> Changing the expiration date will not affect when renewal payments are processed.

I definitely think we should add those! 👍

thomasfw:
What are your thoughts on disabling some of the inputs when a user has an `active` and `recurring` subscription set up inside a payment gateway?

As an example, I can't see any good reason why somebody would want to update a member's subscription expiry, when it doesn't also update the renewal payment date in the gateway. From our experience it only leads to confusion from our site editors, and makes the subscription record in RCP out of sync with the real one in the gateway. The worst part is that it's (currently) not obvious at all that this doesn't update the actual gateway sub, although I can see that's being worked on in this redesign.

I've added this for the current Edit Member screen (2.9.7), and clicking _'Click here to edit this member anyway'_ reloads the page with a `force_edit=true` param in the url, so the user is free to edit as before if they know what they're doing (e.g. they have the ability to log into the gateway and also manually adjust the subscription from there to match):

[...]

PR #1712, opened 2018-03-07
WIP PR: Add quick way to fully delete a member #1177

### LINKED ISSUE #1177: Add quick way to fully delete a member
I get this support question a lot. People have to delete the user account in the "Users" menu, which isn't really very intuitive. It would be nice to have an option for this in the RCP Members table (maybe a bulk action and a row action).

nosegraze:
Closes #1177


includes/admin/admin-notices.php
```diff
@@ -142,6 +142,11 @@ function rcp_admin_notices() {
 
 				$text = __( 'The user\'s email has been verified successfully', 'rcp' );
 				break;
+
+			case 'member_deleted' :
+
+				$text = __( 'Member successfully deleted..', 'rcp' );
```
mindctrl: There's an extra `.` here.

includes/scripts.php
```diff
@@ -44,6 +44,7 @@ function rcp_admin_scripts( $hook ) {
 		wp_localize_script( 'rcp-admin-scripts', 'rcp_vars', array(
 				'rcp_member_nonce'    => wp_create_nonce( 'rcp_member_nonce' ),
 				'cancel_user'         => __( 'Are you sure you wish to cancel this member\'s subscription?', 'rcp' ),
+				'delete_member'       => __( 'Are you sure you want to delete this member? The account and all associated data will be permanently deleted. Payment records will remain but will no longer be associated with an account. Proceed?', 'rcp' ),
```
mindctrl: I think we should say "**user** account" here, just to make it extra clear that we're deleting that too. It may also be worth mentioning that it's irreversible, to reinforce the "Permanently Delete" text.

Regarding https://github.com/restrictcontentpro/restrict-content-pro/issues/1177#issuecomment-294487081, we've had a number of people that just want to remove the membership from the account without deleting the account. I haven't seen a request for it lately though, have you @nosegraze?

nosegraze: @mindctrl I haven't gotten any requests. I just figured I'd hold off on it until we get in the new membership table and maybe the new UI as well ( #1003 ). The former will definitely affect how the code works, and the latter might affect where in the UI we even put that option.

Issue #1645, opened 2018-02-01
Improve reports


nosegraze:
Going to log all customer requests here:

- [ ] Number of subscriptions sold over a certain period. The "Signups" tab shows number of new registered users, but a customer wanted to see total number of subscriptions sold (including upgrades/downgrades) and perhaps with the ability to filter by subscription level.
- [ ] Upgrades over a certain period.
- [ ] Downgrades over a certain period.
- [ ] Cancellations over a certain period.
- [ ] Expirations over a certain period.
- [ ] Renewals (number and dollar amount) over a certain period.
- [ ] Active subscribers over time - see https://github.com/restrictcontentpro/restrict-content-pro/issues/289

tnorthcutt:
Related: currently the refunds report uses the date of the original payment. This is a bit unintuitive, IMO. If I go to a page that says `Refunds Report` and filter for `This Week`, I'd expect to see a refund I just processed for a customer. But if the payment that was refunded happened a month ago, the refund doesn't show up in this report.

Because of how payment data is stored, I don't think we have the ability to query based on when a refund was issued. But, perhaps we could add some language to the refunds report that clarifies what is being shown.

pippinsplugins:
We'll need to create a new "refund" record when a refund is processed. This could be a date and amount column in the existing payments table or a separate table.

tnorthcutt:
@pippinsplugins IMO a separate table makes sense; partial refunds could result in more than one refund occurring for a single payment, which would make storing as columns on the existing payments table complicated.

I'll open a separate issue for better refund recording; obviously this isn't a blocker for improving the reporting overall.

tnorthcutt:
I did some of this for a client as a custom add-on. Our approach there was to create a custom post type for the things we wanted to record (failed payments, upgrades, and cancellations), and save a new post of that type whenever one of those events occurred. We then provide a simple csv export.

For most of the things on the list in the OP, I think we'd likely need to explicitly store the event occurring (upgrade, downgrade, expiration, etc.). Some of them could possibly be pieced together as is, but explicitly storing for reporting purposes would be much cleaner, IMO.

What's the preferred way to do this? Should we have a separate table for many of these? Upgrades and downgrades could probably be combined in "plan changes", with a flag for up/down.

chrisvanpatten:
Could also relate to #1104 in terms of tracking events...

nosegraze:
[...]

PR #1714, opened 2018-03-09
[WIP] Use Chart.js for earnings, refunds, and signups graphs

tnorthcutt:
In progress, not ready for merge consideration 😄


tnorthcutt:
@nosegraze I'd love some additional eyes on this at this point (mostly the earnings graph, as that's what I've focused on so far). It's a tad unwieldy trying to do this the Right Way™ as this file is kind of the antithesis of Don't Repeat Yourself at this point 😁.

Do you think it's worth the time to further refactor this to make it more DRY and extensible for improvements in https://github.com/restrictcontentpro/restrict-content-pro/issues/1645?

Disclaimer: I'm far from done here, still pulling in chartjs and momentjs from cloudflare vs. enqueueing ourselves, etc. Just making sure it's clear I know it's still messy 👌

nosegraze:
I will take a look at this later today. :) 

I will say that EDD is currently undergoing a big overhaul to improve reporting and it may be worth waiting until that's done so we can bring some of that logic and functionality over here.

tnorthcutt:
@nosegraze I had some inspiration strike last night – I think I've figured out a direction for cleaning this up. I'll push some new changes today.

nosegraze:
👍

tnorthcutt:
Ok, this is looking a lot better now. Cleaned up a lot.

I would definitely like a second opinion on https://github.com/restrictcontentpro/restrict-content-pro/pull/1714/commits/f71211258f877ecc9c0c4b371fcef69e7ce2c6ff

This is overall probably ready for consideration. I'm sure some things will need to be tweaked, but it'd probably be overkill for me to try to refine further at this point.

pippinsplugins:
@tnorthcutt I don't see any issues with that commit. Did you have any particular worries with it?

tnorthcutt:
@pippinsplugins the only thing that comes to mind is that it removes filters for time format and tick size. But when I was working on this it seemed like those may have been vestigial by now anyway. Not 100% certain of that... but IMO it makes sense anyway to just handle it correctly in core so that filtering isn't necessary. I'd be really surprised if anyone was previously using those filters.

The only other concern is that while the charts are responsive, the narrow view IMO results in a chart that is a bit shorter than idea. That's a result of the initial canvas height that is set... but setting it taller means a "normal" (e.g. 13" laptop full screen window) view gets too tall, IMO. Open to feedback, even if that's "it's fine, ship it" 😄.

includes/admin/reports/reports-page.php
```diff
@@ -36,246 +36,243 @@ function rcp_reports_page() {
...
- *
- * @access  public
- * @since   1.8
- * @return  void
-*/
-function rcp_earnings_graph() {
-	global $rcp_options, $wpdb;
+ * @return array|float
+ */
+function rcp_earnings_graph_data( $dates, $payments_db, $total, $day_by_day, $subscription, $which = 'earnings', $return_total = false ) {
```
nosegraze: I'm not sure if we really need all these parameters?

- `$dates` - It seems like every time this is passed through it's just `rcp_get_report_dates()`. Can we not just use that inside the function instead of passing it through as a parameter?
- `$payments_db` - We don't need to pass this through. We can just initialize the class inside this function.
- `$total` - It looks like this is always passed through as `0` or `0.00`.
- `$day_by_day` - This seems to always be passed through as `rcp_reports_graph_day_by_day( $dates )` so can we just get that inside the function, just as suggested with `$dates` ?
- `$subscription` - I don't think we need to pass this through either, as it's always just retrieved as `isset( $_GET['subscription'] ) ? absint( $_GET['subscription'] ) : false;`.

includes/admin/reports/reports-page.php
```diff
@@ -36,246 +36,243 @@ function rcp_reports_page() {
...
+
+	if ( $return_total ) {
+		return $total;
+	}
+
+	// data needs to be array of arrays in the form of [YYYY MM DD, amount]
+	return $data;
+}
+
+function rcp_reports_graph_output( $chart_data, $label, $total ) {
```
nosegraze: Can we add PHPDocs here?

includes/admin/reports/reports-page.php
```diff
@@ -36,246 +36,243 @@ function rcp_reports_page() {
...
+	ob_start(); ?>
+    <script>
```
nosegraze: Most of the JavaScript below is indented with spaces - should be tabs.

includes/admin/reports/reports-page.php
```diff
@@ -36,246 +36,243 @@ function rcp_reports_page() {
...
+            var config = {
+                type: 'line',
+                data: {
+                    datasets: [{
+                        label: <?php echo "'$label'"; ?>,
+                        data: [<?php foreach ( $chart_data as $point ) {
+							echo $point[1] . ',';
+						} ?>],
+                        borderColor: '#2A76D2',
+                        backgroundColor: '#2A76D2',
```
nosegraze: It would be really snazzy if we could use the primary colour from the current admin colour scheme.

includes/admin/reports/reports-page.php
```diff
@@ -36,246 +36,243 @@ function rcp_reports_page() {
...
+                <p class="rcp_graph_totals"><strong><?php _e( 'Total ' .  strtolower( $label ) . ' for period shown: ', 'rcp' ); echo $total; ?></strong></p>
+            </div>
+        </div>
+    </div>
+    <?php
+    return ob_get_clean();
+}
 
-	// Determine graph options
+function rcp_reports_graph_day_by_day( $dates ) {
```
nosegraze: PHPDocs.

includes/admin/reports/reports-page.php
```diff
@@ -287,233 +284,17 @@ function rcp_flot_tooltip(x, y, contents) {
...
-	<?php
-	echo ob_get_clean();
+	$dates        = rcp_get_report_dates();
+	$day_by_day   = rcp_reports_graph_day_by_day( $dates );
+	$refunds      = (float) 0.00; // Total refunds for time period shown
+	$subscription = isset( $_GET['subscription'] ) ? absint( $_GET['subscription'] ) : false;
+	$payments_db  = new RCP_Payments;
+	$chart_data   = rcp_earnings_graph_data( $dates, $payments_db, $refunds, $day_by_day, $subscription, 'refunds', false );
+	$refunds      = rcp_earnings_graph_data( $dates, $payments_db, $refunds, $day_by_day, $subscription, 'refunds', true );
+	$total = rcp_currency_filter( $refunds );
```
nosegraze: Variable needs to be lined up.

mindctrl:
@tnorthcutt I think you linked to the wrong commit you were concerned about.

tnorthcutt:
@mindctrl that actually is what I want a second opinion on – is nuking that wrapper div ok? I did that so the controls would stay visible, but not sure if that's the Right Way™ to do so.

includes/admin/reports/reports-page.php
tnorthcutt: 👍

includes/admin/reports/reports-page.php
tnorthcutt: 👍

includes/admin/reports/reports-page.php
tnorthcutt: Thanks. This makes it _much_ cleaner. Not sure why I was blind to those improvements.

includes/admin/reports/reports-page.php
tnorthcutt: 👍

includes/admin/reports/reports-page.php
tnorthcutt: 👍

includes/admin/reports/reports-page.php
tnorthcutt: Oooh, I like this idea! I went with the 4th color in the scheme, mostly because the default scheme results in near-black if we use the primary, which seemed pretty boring. Using a complementary color seemed like a good idea. Obviously that's a minor point and I'm fine with changing it 😄

mindctrl:
@tnorthcutt gotcha. I haven't dug into this and given it focus yet, but at the moment I don't see any issues with removing it.

tnorthcutt:
@nosegraze you good with the adjustments I made in response to your review?

tnorthcutt:
I think this may be ready for `3.0`, though more eyes won't hurt.

Issue #1196, opened 2017-04-21
Validate payment gateway API keys


tnorthcutt:
Recently, a client inadvertently pasted some extra text in along with his Stripe API key, including some whitespace. That resulted in this in the devtools console:

![image](https://cloud.githubusercontent.com/assets/796639/25283545/5f08ed0c-2671-11e7-9376-93001b591ce1.png)

This (at least when the extra characters include whitespace) can be avoided in the future if we validate payment gateway API key fields, either in PHP when saved, or preferably in javascript before saving.

I'm opening this issue for posterity; someone from my team may be able to PR this in the near future but otherwise I thought I'd at least note it here in case anyone else wants to tackle it.

nosegraze:
I had a customer include `#` at the beginning of their 2Checkout merchant ID number, which broke the whole payment process.

pippinsplugins:
>>>For Stripe, I made it very strict, since they follow (AFAIK) a very strict pattern of sk_live_xxxxx, sk_test_xxxxx, etc.

@tnorthcutt unfortunately that's not quite true. API keys from old Stripe accounts (such as mine) do not follow that pattern. My test Stripe secret starts with `SduaJk`.

tnorthcutt:
@pippinsplugins Ah, good to know! I'll update to only require numbers/letters/underscores.

PR #1725, opened 2018-03-17
Validate payment gateway API keys

tnorthcutt:
As of now this only includes Stripe and PayPal – and PayPal is pretty loose; it only requires that there be no spaces. I did it that way since there are so many variations of PayPal products and platforms, and I wasn't sure how strict we should get.

For Stripe, I made it very strict, since they follow (AFAIK) a very strict pattern of sk_live_xxxxx, sk_test_xxxxx, etc.

I'm not as familiar with other gateways; I'd love some input from others on what patterns we should use for them.

To keep it simple, I used `title` attributes to customize the validation error messaging. Of course if we wanted to we could get fancy with custom callbacks, determining what validation rule exactly failed (e.g. a space was added, a `#` character, etc.).

But for now, IMO shipping > perfection, especially with a small feature like this.

Closes https://github.com/restrictcontentpro/restrict-content-pro/issues/1196


tnorthcutt:
@nosegraze 👍 thanks

nosegraze:
The PayPal ones aren't working for me. I have my credentials entered but it's throwing the error to check the input.

tnorthcutt:
@nosegraze thank you for the heads up. Can you post an example string that is similar to your Paypal credentials? (e.g. use different characters, just the same pattern, for security)

nosegraze:
PayPal Address: `ashleygibson@nosegraze.com`
Test API Username: `ashleygibson_api1.nosegraze.com`
Test API Password: `ABCDEFG123` (anything with numbers/letters)
Test API Signature: `ABcDefg123a` (same as above - anything with numbers/letters)

There aren't any spaces in any of them. I basically can't get anything to save.

I tried saving just `a` in the signature field and that worked. Just saving `1` worked. But saving `a1` did not work.

tnorthcutt:
@nosegraze thanks. Think I've got it sorted now.

chrisvanpatten:
Potential devil's advocate/counterpoint…

I spent a while buidling a SaaS service at one point that had a lot of API integrations with 3rd party services, so validating the API credentials was a pretty big part of the UX.

We started by doing API key validation via regex, but over time the edge cases became really hard to keep track of, and basically required us to dedicate humanpower (and a lengthy Google Doc) to tracking potential key formats for each service.

Eventually, we switched to a sort of integration testing-style method. We added "validate_connection" methods to every integration, which was run when saving the API credentials. It would make a small read-only request to each service's API (for example for Stripe you might make a request to the `GET https://api.stripe.com/v1/plans` endpoint; some other services have "account" or "meta" endpoints you can hit that work even better). If the request was successful, you're good to go. If not, you get a nice error message clearly indicating the problem, and know the credentials were wrong.

This worked even better than RegEx testing because (a) you don't have to worry about tracking changes to API key formats and (b) you actually validate that it's a valid key, not just a set of characters that happen to match the right format.

Obviously this means modifying the payment gateways to support such a feature, which is more involved up front, but it might be worth considering!

mindctrl:
I like @chrisvanpatten's idea.

It's also worth pointing out we're hooking into Stripe Connect in 2.9.11, so adding any additional validation to that would be counterproductive and confusing if the regex validation fails after a successful connection. Also the Stripe API key fields are going to be hidden, so the errors on those wouldn't be visible, and the inability to submit the settings form would be even more confusing.

tnorthcutt:
@chrisvanpatten very nice idea. I'd originally considered that and dismissed it as more complicated than necessary, but on reflection it probably is the most user-helpful approach. Thanks for bringing it up!

@mindctrl should we push pause on this until the Stripe Connect integration is finalized?

mindctrl:
@tnorthcutt yes, let's hold off. We have Stripe Connect in, but I think in general this approach for any of the gateways brings the possibility of the issues outlined above. I'm going to close this PR. We can tackle this another way entirely in a new one.

tnorthcutt:
@mindctrl Sounds good to me.

PR #1726, opened 2018-03-17
Discount codes: allow multi select for subscription level

### LINKED ISSUE #1543: Discount codes: allow multi select for subscription level
Got a request to allow a discount code to be applied to multiple specific subscription levels instead of just one.

tnorthcutt:
Heavily inspired by EDD.

This does have the potential to break back-compat for any 3rd party devs using functions which have been removed. I'm open to thoughts on whether/how we should approach that.

Closes #1543 

_N.B. I expect some tests to fail; I left old ones in to ensure they do fail as expected before removing, as a sanity check. This PR also includes new tests._


tnorthcutt:
Ah, the tests I expected to fail did fail (the ones for the old single-subscription discount codes).

But, the new ones also failed, because the test setup (apparently?) did not run the upgrade routine. Not sure what the procedure is for that, though. For now in the upgrade routine I assumed this will get merged for 3.0, but I don't know how to get the test setup to recognize that and run the upgrade.

includes/admin/class-rcp-upgrades.php
```diff
@@ -139,6 +140,44 @@ private function v29_upgrades() {
...
+
+		if ( version_compare( $this->version, '3.0', '<' ) ) {
+
+			global $wpdb;
+			$class = new RCP_Discounts();
+
+			rcp_log( 'Performing version 3.0 upgrades.', true );
+
+			$prefix = $wpdb->prefix;
+			$check  = $wpdb->get_row( "SELECT * FROM {$prefix}rcp_discounts" );
```
nosegraze: All instances of `{$prefix}rcp_discounts` should be `rcp_get_discounts_db_name()` instead.

includes/admin/discounts/discount-codes.php
```diff
@@ -98,12 +98,15 @@ function rcp_discounts_page() {
 							</td>
 							<td data-colname="<?php _e( 'Description', 'rcp' ); ?>"><?php echo stripslashes( $code->description ); ?></td>
 							<td data-colname="<?php _e( 'Code', 'rcp' ); ?>"><?php echo $code->code; ?></td>
-							<td>
+							<td data-colname="<?php _e( 'Subscription', 'rcp' ); ?>">
```
nosegraze: This should be `esc_attr_e()` (I know the previous ones don't use that correctly, but really it should be `esc_attr_e()`). Maybe you could update the others while you're in there. :)

includes/admin/discounts/edit-discount.php
```diff
@@ -69,22 +69,28 @@
...
+					<label for="rcp-subscriptions"><?php _e( 'Subscriptions', 'rcp' ); ?></label>
 				</th>
 				<td>
 					<?php
-					$levels = rcp_get_subscription_levels('all', false);
+					$levels = rcp_get_subscription_levels('all' );
 					if( $levels ) : ?>
-						<select name="subscription" id="rcp-subscription">
-							<option value="0"><?php _e( 'All Levels', 'rcp' ); ?></option>
+                        <?php
```
nosegraze: Formatting is off here.

nosegraze:
I left a few comments just regard to code style stuff. I haven't actually tested this yet or anything. A few other things:

- This PR should be against the `release/3.0` branch. Once changed, that will probably create a merge conflict, as we already have a `3.0` upgrade routine. So that will need to be adjusted.
- Would it be better to store the subscription level IDs as a comma-separated string rather than a serialized array?
- We should keep the `get_subscription_id()` and `has_subscription_id()` methods for backwards compatibility and perhaps just return the first selected subscription ID.

nosegraze:
I like "membership levels". I see that terminology in pre-sales tickets quite a bit. People will ask if they can _"create multiple membership levels"_ and things like that.

mindctrl:
👍 "membership levels" works for me too.

includes/admin/class-rcp-upgrades.php
```diff
@@ -165,11 +165,40 @@ private function v30_upgrades() {
...
+			// Upgrade discounts table
+			$class = new RCP_Discounts();
+
+			rcp_log( 'Performing version 3.0 upgrades.', true );
+
+			$discounts_table = rcp_get_discounts_db_name();
+			$check  = $wpdb->get_row( "SELECT * FROM {$discounts_table}" );
+
+			if ( ! isset( $check->subscription_ids ) ) {
+				$updated = $wpdb->query( "ALTER TABLE {$discounts_table} ADD subscription_ids LONGTEXT NOT NULL" );
```
mindctrl: @tnorthcutt Is there any particular reason to use `LONGTEXT` here? It uses more space and memory and I don't think we'll ever need anywhere near 4GB of storage space for plan IDs.

includes/admin/discounts/discount-codes.php
tnorthcutt: 👍
Also changing to `esc_html_e()` as appropriate.

includes/admin/class-rcp-upgrades.php
tnorthcutt: @mindctrl no, good catch. Thanks. Changing to `TEXT`.

tnorthcutt:
> This PR should be against the release/3.0 branch. Once changed, that will probably create a merge conflict, as we already have a 3.0 upgrade routine. So that will need to be adjusted.

Done!

> Would it be better to store the subscription level IDs as a comma-separated string rather than a serialized array?

I'm almost certain serialized is the way to go, since WP includes magic `maybe_serialize()` and `maybe_unserialize()` functions for handling the data. Otherwise we end up parsing the comma-separated string ourselves and have to guard against improper formatting.

> We should keep the get_subscription_id() and has_subscription_id() methods for backwards compatibility and perhaps just return the first selected subscription ID.

Done in https://github.com/restrictcontentpro/restrict-content-pro/pull/1726/commits/a237cc6903bf6939ff6289f3a6324aa49b45629e

tnorthcutt:
@nosegraze @mindctrl re: `subscriptions` vs. `membership levels`, are you talking solely about the UI, or also about naming in code? Happy to make the change; just want to check how extensive to be.

nosegraze:
> @nosegraze @mindctrl re: subscriptions vs. membership levels, are you talking solely about the UI, or also about naming in code? Happy to make the change; just want to check how extensive to be.

I reckon both. We won't edit anything that already there that would break backwards compatibility, but for all our new code going forward we should start moving to using `membership levels`. We'll probably be adding a ton of new stuff when we add the new table and support for multiple subscriptions, and we can work on making all those new functions/classes use the `membership` wording instead.

tnorthcutt:
@nosegraze cool, makes sense. I'll work on changing language for all newly introduced code as appropriate.

tnorthcutt:
Ok, this looks better now. Ready for reviewing 👍

tnorthcutt:
Any chance of getting this merged into `release/3.0` soon? https://github.com/restrictcontentpro/restrict-content-pro/issues/1745 is also going to require [Chosen](https://harvesthq.github.io/chosen/), so it'd be nice to have that already in the release branch and avoid the messy merge of adding it in two different branches.

mindctrl:
@tnorthcutt I'd rather we use standard checkboxes instead of Chosen. One reason why is the WP core team is hoping to get Select2 in core, as per https://github.com/select2/select2/issues/3744.

I also don't think most sites have a ton of levels, so a checkbox group wouldn't be huge on most sites.

Will review this PR by mid week.

tnorthcutt:
@mindctrl the issue you linked to has been open for 2.5 years, and the corresponding [trac ticket](https://core.trac.wordpress.org/ticket/31696) for three years, with no movement in the past 10 months.

If Select2 is indeed a better choice, then by all means let's use that, but I don't think holding our breath for core including it is going to pay off.

mindctrl:
@tnorthcutt is there a compelling reason to use it over standard checkboxes? Since most sites we see have only a handful of levels, I don't think we need a fancy dropdown to make a couple selections when a few checkboxes will do. Also EDD uses Chosen and we've seen numerous quirks with it there.

tnorthcutt:
@mindctrl roger that; I did not know about the Chosen issues in EDD. I modeled this after EDD, which is why I _chose_ Chosen, erroneously thinking it'd be "tried and tested" since EDD has been using it.

pippinsplugins:
I vote for checkboxes as well.

tnorthcutt:
Checkboxes it is 😁

nosegraze:
Yes I'd prefer checkboxes too. We already use them in the restrictions interface without issue. 👍

tnorthcutt:
_Ignore this for now; I pushed to the wrong branch. Cleaning up._

includes/admin/class-rcp-upgrades.php
```diff
@@ -165,11 +165,40 @@ private function v30_upgrades() {
...
+			$check  = $wpdb->get_row( "SELECT * FROM {$discounts_table}" );
+
+			if ( ! isset( $check->membership_level_ids ) ) {
+				$updated = $wpdb->query( "ALTER TABLE {$discounts_table} ADD membership_level_ids TEXT NOT NULL" );
+
+				// Migrate existing subscriptions to new column in proper format
+				$discounts = rcp_get_discounts();
+				foreach ( $discounts as $discount ) {
+					$membership_level_ids = ( $discount->subscription_id == 0 ) ? array() : array( $discount->subscription_id );
+					$args             = array(
```
nosegraze: Variable not quite lined up with the one above.

includes/admin/class-rcp-upgrades.php
```diff
@@ -165,11 +165,40 @@ private function v30_upgrades() {
 
 			rcp_log( sprintf( 'Renaming payment_id to rcp_payment_id in %s was successful.', $table_name ), true );
 
-			$this->upgraded = true;
-		}
-	}
+			// Upgrade discounts table
+			$class = new RCP_Discounts();
```
nosegraze: For clarity, can this variable please be renamed to `$discounts` or similar?

includes/admin/discounts/discount-actions.php
```diff
@@ -32,15 +32,15 @@ function rcp_process_add_discount() {
...
-		'subscription_id' => $_POST['subscription']
+		'name'             => $_POST['name'],
+		'description'      => $_POST['description'],
+		'amount'           => $_POST['amount'],
+		'unit'             => isset( $_POST['unit'] ) && $_POST['unit'] == '%' ? '%' : 'flat',
+		'code'             => $_POST['code'],
+		'status'           => 'active',
+		'expiration'       => $_POST['expiration'],
+		'max_uses'         => $_POST['max'],
+		'membership_level_ids' => isset( $_POST['membership_levels'] ) ? $_POST['membership_levels'] : array()
```
nosegraze: Can we line up all the array keys please?

includes/admin/discounts/discount-actions.php
```diff
@@ -84,15 +84,15 @@ function rcp_process_edit_discount() {
...
-		'subscription_id' => $_POST['subscription']
+		'name'             => $_POST['name'],
+		'description'      => $_POST['description'],
+		'amount'           => $_POST['amount'],
+		'unit'             => isset( $_POST['unit'] ) && $_POST['unit'] == '%' ? '%' : 'flat',
+		'code'             => $_POST['code'],
+		'status'           => $_POST['status'],
+		'expiration'       => $_POST['expiration'],
+		'max_uses'         => $_POST['max'],
+		'membership_level_ids' => isset( $_POST['membership_levels'] ) ? $_POST['membership_levels'] : array()
```
nosegraze: Can we line up all the array key values please?

includes/admin/discounts/discount-codes.php
```diff
@@ -60,16 +60,16 @@ function rcp_discounts_page() {
...
-						<th scope="col" class="rcp-discounts-amount-col"><?php _e( 'Amount', 'rcp' ); ?></th>
-						<th scope="col" class="rcp-discounts-type-col"><?php _e( 'Type', 'rcp' ); ?></th>
-						<th scope="col" class="rcp-discounts-status-col"><?php _e( 'Status', 'rcp' ); ?></th>
-						<th scope="col" class="rcp-discounts-uses-col"><?php _e( 'Uses', 'rcp' ); ?></th>
-						<th scope="col" class="rcp-discounts-uses-left-col"><?php _e( 'Uses Left', 'rcp' ); ?></th>
-						<th scope="col" class="rcp-discounts-expir-col" ><?php _e( 'Expiration', 'rcp' ); ?></th>
+						<th scope="col" class="rcp-discounts-name-col column-primary" ><?php esc_html_e( 'Name', 'rcp' ); ?></th>
+						<th scope="col" class="rcp-discounts-desc-col"><?php esc_html_e( 'Description', 'rcp' ); ?></th>
+						<th scope="col" class="rcp-discounts-code-col" ><?php esc_html_e( 'Code', 'rcp' ); ?></th>
+						<th scope="col" class="rcp-discounts-membership-levels-col" ><?php esc_html_e( 'Membership Levels', 'rcp' ); ?></th>
```
nosegraze: Maybe we should keep the old class name here for backwards compatibility in case someone used it for styling. So it should be:

`class="rcp-discounts-membership-levels-col rcp-discounts-subscription-col"`

includes/admin/discounts/discount-codes.php
```diff
@@ -118,132 +121,130 @@ function rcp_discounts_page() {
...
```
nosegraze: Same as the heading. Keep the old class name in addition for backwards compatibility.

includes/admin/discounts/discount-codes.php
```diff
@@ -118,132 +121,130 @@ function rcp_discounts_page() {
...
-												foreach( $levels as $level ) :
-													echo '<option value="' . $level->id . '">' . $level->name . '</option>';
-												endforeach;
-											?>
-										</select>
-									<?php endif; ?>
-									<p class="description"><?php _e( 'The subscription levels this discount code can be used for.', 'rcp' ); ?></p>
+									if ( $levels ) :
+										foreach ( $levels as $level ) : ?>
+											<input type="checkbox" id="rcp-membership-levels-<?php echo $level->id; ?>" name="rcp_membership_levels[]" value="<?php echo $level->id ?>">
```
nosegraze: Please put the level ID in `esc_attr()` in both the `id` and `value` attributes.

includes/admin/discounts/discount-codes.php
```diff
@@ -118,132 +121,130 @@ function rcp_discounts_page() {
...
+											<label for="rcp-membership-levels-<?php echo $level->id; ?>"><?php echo $level->name; ?></label>
```
nosegraze: `esc_attr()` on the ID in the label, and `esc_html()` on the name.

includes/admin/discounts/edit-discount.php
```diff
@@ -69,22 +69,24 @@
...
-									echo '<option value="' . $level->id . '" ' . selected( $code->subscription_id, $level->id, false ) . '>' . $level->name . '</option>';
-								endforeach;
-							?>
-						</select>
-					<?php endif; ?>
-					<p class="description"><?php _e( 'The subscription levels this discount code can be used for.', 'rcp' ); ?></p>
+					$levels = rcp_get_subscription_levels( 'all' );
+					if ( $levels ) :
+						$current = maybe_unserialize( $code->membership_level_ids );
+						if ( ! isset( $current ) || ( is_array( $current ) && empty( $current ) ) || ! $current ) {
```
nosegraze: I think the checks here are a bit excessive. Could probably be simplified with:

```php
if ( empty( $current ) || ! is_array( $current ) )
```

includes/admin/discounts/edit-discount.php
```diff
@@ -69,22 +69,24 @@
...
+							$current = array();
+						}
+						foreach ( $levels as $level ) : ?>
+							<input type="checkbox" id="rcp-membership-levels-<?php echo $level->id; ?>" name="membership_levels[]" value="<?php echo $level->id ?>" <?php checked( true, in_array( $level->id, $current ) ); ?>>
+							<label for="rcp-membership-levels-<?php echo $level->id; ?>"><?php echo $level->name; ?></label>
```
nosegraze: Please add the following:

`esc_attr()` on the `<input>` tag `id` and `value` attributes, and the `<label>` `for` attribute.
`esc_html()` on the level name.

tests/test-discounts.php
```diff
@@ -114,6 +114,14 @@ function test_has_subscription_id() {
 		$this->assertFalse( $this->db->has_subscription_id( $this->discount_id ) );
 	}
 
+	function test_get_membership_level_ids() {
```
nosegraze: Can you please add PHPDocs for these two new tests? I know the old ones don't have them but we should start adding them for any new ones.

includes/admin/class-rcp-upgrades.php
tnorthcutt: 👍

includes/admin/class-rcp-upgrades.php
tnorthcutt: 👍 going with `$discounts_class` (since `$discounts` is used later to, naturally, describe a collection of discounts)

includes/admin/discounts/discount-actions.php
tnorthcutt: 👍

includes/admin/discounts/discount-actions.php
tnorthcutt: 👍

includes/admin/discounts/discount-codes.php
tnorthcutt: Good call!

includes/admin/discounts/discount-codes.php
tnorthcutt: 👍

includes/admin/discounts/discount-codes.php
tnorthcutt: 👍

includes/admin/discounts/discount-codes.php
tnorthcutt: 👍

includes/admin/discounts/edit-discount.php
tnorthcutt: Yep, I like that a lot better. Thank you.

includes/admin/discounts/edit-discount.php
tnorthcutt: 👍

tests/test-discounts.php
tnorthcutt: Absolutely.

tnorthcutt:
@nosegraze Thanks for the very thorough review! I'll try to get better about the tedious little things; I know those are annoying 😕

tnorthcutt:
I believe this is ready for `3.0`, though getting a few more eyes on it wouldn't hurt.

includes/admin/class-rcp-upgrades.php
```diff
@@ -165,11 +165,40 @@ private function v30_upgrades() {
...
-			$this->upgraded = true;
-		}
-	}
+			// Upgrade discounts table
+			$discounts_class = new RCP_Discounts();
+
+			rcp_log( 'Performing version 3.0 upgrades.', true );
+
+			$discounts_table = rcp_get_discounts_db_name();
+			$check  = $wpdb->get_row( "SELECT * FROM {$discounts_table}" );
```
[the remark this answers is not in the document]
tnorthcutt: 👍

PR #1734, opened 2018-03-23
Only disable WP toolbar for members (subscriber role)

### LINKED ISSUE #1729: Adjust description/functionality for "Disable WordPress Toolbar" setting
Right now it says:

> Check this if you'd like to disable the WordPress toolbar for subscribers. Note: will not disable the toolbar for administrators.

First of all, I don't like the usage of "subscribers". I think that word is confusing and we should start trying to move away from it and use "members" instead.

However, either way the description isn't really accurate for what the option does. The setting disables the toolbar for _non-admins_, regardless of whether they have a membership or not.

This confused one customer who thought the toolbar would only be disabled for the `subscriber` role. He was confused as to why it also disappeared for his users in the `editor` role.

We should do one or both of the following:

1. Consider: should we really disable the toolbar for users who are able to edit/create posts (or in general perform any admin action beyond just editing their profile)?
2. We should reword the description to more accurately describe who is going to lose toolbar access.

tnorthcutt:
Also improves wording on settings page.

Closes #1729

### DIFF
--- a/includes/member-functions.php
+++ b/includes/member-functions.php
@@ -1318,11 +1318,7 @@ function rcp_maybe_disable_toolbar() {
 
 	global $rcp_options;
 
-	$user = wp_get_current_user();
-	$roles = ( array ) $user->roles;
-	$role = $roles[0];
-
-	if ( isset( $rcp_options['disable_toolbar'] ) && $role == 'subscriber' ) {
+	if ( isset( $rcp_options['disable_toolbar'] ) && ! current_user_can( 'edit_posts' ) ) {
 		add_filter( 'show_admin_bar', '__return_false' );
 	}
 }
--- a/includes/member-functions.php
+++ b/includes/member-functions.php
@@ -1318,7 +1318,11 @@ function rcp_maybe_disable_toolbar() {
 
 	global $rcp_options;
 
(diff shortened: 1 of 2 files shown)


includes/member-functions.php
```diff
@@ -1318,7 +1318,11 @@ function rcp_maybe_disable_toolbar() {
 
 	global $rcp_options;
 
-	if ( isset( $rcp_options['disable_toolbar'] ) && ! current_user_can( 'manage_options' ) ) {
+	$user = wp_get_current_user();
+	$roles = ( array ) $user->roles;
+	$role = $roles[0];
+
+	if ( isset( $rcp_options['disable_toolbar'] ) && $role == 'subscriber' ) {
```
nosegraze: We can't do the `subscriber` role only because not everyone uses that for their members. Instead we should just check to see if they have a membership level at all ( `rcp_get_subscription_id()` ) and/or remove it for anyone who doesn't have the `edit_posts` capability.

tnorthcutt: I'm going to go with the capability check; it's quite possible that an author/editor/etc. would also have a subscription, whether for testing purposes or actually purchased.

tnorthcutt:
I believe this is ready for merge into `3.0`.

PR #1739, opened 2018-03-28
Register meta table names when switching sites on multisite

### LINKED ISSUE #1547: Register meta table names when switching sites on multisite
It's possible under certain conditions for the `levelmeta` and `paymentmeta` table names to be incorrect on multisite when using `switch_to_blog()`.

From Slack conversation:

>Travis has an interesting issue on multisite.
>
>He's hooking into `rcp_site_creation_site_created` to do some things. One of those things is to create new RCP levels on the new site, and add some level metadata. Level metadata in this case is related to Group Accounts.
>
>So he used `rcpga_set_level_group_seats_allowed()` to update the seat count allowed. This is where the issue comes up.
>
>`rcpga_set_level_group_seats_allowed()` uses the `$rcp_levels_db` global when calling `update_meta()`. `RCP_Levels::update_meta()` uses core's `update_metadata()` function, which itself uses the `$wpdb` global when determining the table name for the passed type. 
>
>We register `$wpdb->levelmeta` in `rcp_register_databases()`, but at no point is that reference updated when switching sites (he is using `switch_to_blog()`). The `$wpdb` global used by `update_metadata()` has the wrong/unprefixed reference to levelmeta.
>
>There's a `switch_blog` hook that we could maybe use to run `rcp_register_databases()` when switching sites. Not sure if there would be any repercussions.
>
>There are probably other cases where this global issue shows its head, so it would be nice to see if that's a solution or if there's another way.

To help spot any possible unintended side effects of doing this, @tnorthcutt is going to be running `rcp_register_databases()` on the `switch_blog` hook while working on his project.

tnorthcutt:
Closes #1547

Signed-off-by: Travis Northcutt <travis@travisnorthcutt.com>


tnorthcutt:
@mindctrl yep! We accidentally removed it with an RCP update... and it'd been long enough that I couldn't remember what the bug was. Eventually we sorted out that we needed to reapply this patch 😂 😭 . All good since re-applying.

PR #1740, opened 2018-03-28
Make admin emails fully filterable

### LINKED ISSUE #1497: Make admin emails filterable
Currently, admin emails are not always filterable. In [`email-functions.php`](https://github.com/restrictcontentpro/restrict-content-pro/blob/724c7c207f2ed022e6daf34edc44a5904268ac49/includes/email-functions.php#L59):

```
$admin_message = isset( $rcp_options['active_email_admin'] ) ? $rcp_options['active_email_admin'] : '';
```

So if an admin message is set, it is not filterable. However, if it is not set, it is now filterable:
```
if( empty( $admin_message ) ) {
    $admin_message = __( 'Hello', 'rcp' ) . "\n\n" . $user_info->display_name . ' (' . $user_info->user_login . ') ' . __( 'is now subscribed to', 'rcp' ) . ' ' . $site_name . ".\n\n" . __( 'Subscription level', 'rcp' ) . ': ' . rcp_get_subscription( $user_id ) . "\n\n";
    $admin_message = apply_filters( 'rcp_before_admin_email_active_thanks', $admin_message, $user_id );
    $admin_message .= __( 'Thank you', 'rcp' );
}
```

I propose adding filters, like so:
```
$admin_message = isset( $rcp_options['active_email_admin'] ) ? $rcp_options['active_email_admin'] : '';
$admin_message = apply_filters( 'rcp_subscription_active_admin_email', $admin_message, $user_id, $status );
```

And the same for the subject, and for each status (`active`, `cancelled`, `expired`, `free`, `trial`).

If this is something that would be merged, give me the thumbs up and I'll PR it ASAP.

_My use case, FWIW: a client wants notification emails to specify whether a registration is for a new user or for a plan change. Currently, I can't see a way to accomplish this, without these filters. Any suggestions for a workaround are of course welcome._

tnorthcutt:
Closes #1497


nosegraze:
> The thinking there is that eventually all our filters would have a natural grouping... rcp_email_ in the case of emails.
> 
> Thoughts?

I like that. 👍

tnorthcutt:
I like that natural grouping as well. My only hesitation is that `rcp_subscription_cancelled_admin_email_message` is "readable", if that makes sense.

"What is this?"

"It's the RCP filter for the subscription cancelled admin email message."

As opposed to:

"It's the RCP filter for the email admin subscription cancelled message."

¯\\\_(ツ)_/¯

tnorthcutt:
@mindctrl @nosegraze any feedback on my last comment re: filter naming?

mindctrl:
I care more about nice categorization of the type of filter it is than I do about the way it may read as a sentence. One benefit is it's easier to search code and docs for `rcp_email_` and find all the email filters, and also all the places that filter may be used in RCP and in other plugins. And when documenting them for public use, they're easier to read and sort if they're prefixed the same.

nosegraze:
I agree with @mindctrl .

nosegraze:
I just had a thought: do we want to use `membership` instead of `subscription`? It would be inconsistent with what we already have, but would be thinking more towards the future when we start renaming things.

tnorthcutt:
@nosegraze I guess that depends: once we have multiple subscriptions possible, will these emails (for instance the `active` email) be sent when someone becomes a member for the first time, or will they be sent any time someone activates a new subscription (or cancels one, etc.)?

mindctrl:
I think they would be sent for each subscription.

nosegraze:
Agreed. The email is basically another confirmation of _"everything went well during registration"_ or _"this specific membership has expired"_ (it wouldn't make sense for that to only be sent for one membership but not the other).

tnorthcutt:
@nosegraze just pushed a change to the filter names 👍

nosegraze:
You happy with this naming, @mindctrl ?

mindctrl:
@tnorthcutt I think I'd prefer a slight adjustment.

instead of:
`rcp_email_admin_subject_membership_expired`
`rcp_email_admin_message_membership_expired`

how about:
`rcp_email_admin_membership_expired_subject`
`rcp_email_admin_membership_expired_message`

That makes the character group idential to the very last thing.

tnorthcutt:
@mindctrl 👍 got those changed

mindctrl:
I'm happy with this now. @nosegraze you?

nosegraze:
Yes I'm happy. :+1:

PR #1741, opened 2018-03-29
Identify RCP pages in the list table screen

### LINKED ISSUE #1617: Identify RCP pages in the list table screen
Some plugins, like WooCommerce, identify their pages in the Pages > All Pages list view.

Example, note the "Cart Page" and "Checkout Page" hints after the linked title:
image

I find this a useful hint when logged into customer sites, and think it could be useful for us to do the same.

We can use the `display_post_states` filter to do this.

tnorthcutt:
Closes #1617

### DIFF
--- a/includes/admin/admin-actions.php
+++ b/includes/admin/admin-actions.php
@@ -39,7 +39,7 @@ function rcp_display_post_states( $post_states, $post ) {
 	$rcp_options = get_option( 'rcp_settings' );
 	$pages       = array(
 		'registration_page' => 'Registration Page',
-		'redirect'          => 'Redirect Page',
+		'redirect'          => 'Registration Success Page',
 		'account_page'      => 'Account Page',
 		'edit_profile'      => 'Edit Profile Page',
 		'update_card'       => 'Update Card Page'
--- a/includes/admin/admin-actions.php
+++ b/includes/admin/admin-actions.php
@@ -28,3 +28,31 @@ function rcp_process_actions() {
 	}
 }
 add_action( 'admin_init', 'rcp_process_actions' );
+
+/**
+ * Denotes the various RCP pages as such in the pages list table.
+ *
+ * @param array $post_states An array of post display states.
+ * @param WP_Post $post The current post object.
+ */
+function rcp_display_post_states( $post_states, $post ) {
+	$rcp_options = get_option( 'rcp_settings' );
+	$pages       = array(
+		'registration_page' => 'Registration Page',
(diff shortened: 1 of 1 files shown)


includes/admin/admin-actions.php
```diff
@@ -27,4 +27,32 @@ function rcp_process_actions() {
...
+ * Denotes the various RCP pages as such in the pages list table.
+ *
+ * @param array $post_states An array of post display states.
+ * @param WP_Post $post The current post object.
+ */
+function rcp_display_post_states( $post_states, $post ) {
+	$rcp_options = get_option( 'rcp_settings' );
+	$pages       = array(
+		'registration_page' => 'Registration Page',
+		'redirect'          => 'Redirect Page',
```
pippinsplugins: I think this should be `Registration Success Page`.

includes/admin/admin-actions.php
```diff
@@ -27,4 +27,32 @@ function rcp_process_actions() {
...
+ */
+function rcp_display_post_states( $post_states, $post ) {
+	$rcp_options = get_option( 'rcp_settings' );
+	$pages       = array(
+		'registration_page' => 'Registration Page',
+		'redirect'          => 'Registration Success Page',
+		'account_page'      => 'Account Page',
+		'edit_profile'      => 'Edit Profile Page',
+		'update_card'       => 'Update Card Page'
+	);
```
mindctrl: I think these page names need to be translatable.

tnorthcutt: Good point, thanks!

includes/admin/admin-actions.php
```diff
@@ -27,4 +27,32 @@ function rcp_process_actions() {
...
-add_action( 'admin_init', 'rcp_process_actions' );
\ No newline at end of file
+add_action( 'admin_init', 'rcp_process_actions' );
+
+/**
+ * Denotes the various RCP pages as such in the pages list table.
+ *
+ * @param array $post_states An array of post display states.
+ * @param WP_Post $post The current post object.
+ */
```
nosegraze: Can you please add `@return` to the PHPDocs?

tnorthcutt: Err, it got merged before I got a chance to do that – what's your preference on how to do that now?

mindctrl: I fixed it before merging.

tnorthcutt: alright-alright-alright

Issue #1621, opened 2018-01-06
Registration Form UX Issues


thomasfw:
1. On pressing the `enter` key when an input has focus, the  _"Please Wait..."_ overlay is displayed but the form is not submitted. This only seems to happen when the selected subscription is a **paid** one.

2. When errors are displayed, can we auto-scroll to them? Or maybe add a styled 'rcp-input-error' class to each of the invalid fields instead. It's easy to miss the error message at the bottom of the form.

nosegraze:
Regarding your second point, see https://github.com/restrictcontentpro/restrict-content-pro/issues/1394 (as that will probably be required or easy to add when we fix that issue).

nosegraze:
A few items in this related issue https://github.com/restrictcontentpro/restrict-content-pro/issues/1394 weren't addressed so I'm copying them over here:

> When errors are present, in addition to the error message above the submit, an error message should appear in the label of the input that contains an error.
> 
> When submit button is pressed without filling out any information, no individual fields are in error but should be.

PR #1742, opened 2018-03-29
Improve accessibility of registration form errors

### LINKED ISSUE #1394: Accessibility issues
A customer emailed these suggestions to us to fix some accessibility issues on the registration page.

> First item: 

> The list of form error messages is represented in <p> paragraph elements. For a11y, these should be semantically represented in a <ul> unordered list. Or, if the elements cannot be changed, use ARIA roles:

> `<div class="rcp_message" role="list">`
> `<p class="rcp_error" role="listitem">`

> Second item:

> Since the error messages are dynamically added to the page without refresh, the "rcp_message error" div needs to receive focus so that screen readers go straight to the error message and read it instead of waiting on the submit button with no indicators anything on page has changed.

> When errors are present, in addition to the error message above the submit, an error message should appear in the label of the input that contains an error.

> When submit button is pressed without filling out any information, no individual fields are in error but should be.

tnorthcutt:
Adds ARIA `list` and `listitem` roles to error message wrapper and individual messages, respectively.

Also focuses the error message div if it appears.

Closes #1394

Partially addresses #1621

### DIFF
--- a/includes/js/register.js
+++ b/includes/js/register.js
@@ -105,7 +105,7 @@ jQuery(document).ready(function($) {
 				$('.rcp_message.error').attr("tabindex",-1).focus();
 			}, 10);
 		}).fail(function( response ) {
-            console.log(response);
+			console.log(response);
 		}).always(function( response ) {
 		});
 
--- a/includes/js/register.js
+++ b/includes/js/register.js
@@ -101,11 +101,10 @@ jQuery(document).ready(function($) {
 
 		}).success(function( response ) {
 		}).done(function( response ) {
-            window.setTimeout(function() {
-                $('.rcp_message.error').attr("tabindex",-1).focus();
-            }, 10);
+			window.setTimeout(function() {
+				$('.rcp_message.error').attr("tabindex",-1).focus();
+			}, 10);
 		}).fail(function( response ) {
-			console.log( response );
 		}).always(function( response ) {
 		});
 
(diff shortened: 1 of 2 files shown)


includes/js/register.js
```diff
@@ -101,6 +101,9 @@ jQuery(document).ready(function($) {
 
 		}).success(function( response ) {
 		}).done(function( response ) {
+            window.setTimeout(function() {
```
pippinsplugins: Indention isn't quite right here.

includes/js/register.js
```diff
@@ -101,8 +101,10 @@ jQuery(document).ready(function($) {
 
 		}).success(function( response ) {
 		}).done(function( response ) {
+			window.setTimeout(function() {
+				$('.rcp_message.error').attr("tabindex",-1).focus();
+			}, 10);
 		}).fail(function( response ) {
-			console.log( response );
```
nosegraze: Why was this removed?

tnorthcutt: Oops – I thought I'd introduced it in a previous commit. Do we actually want to leave a `console.log()` call running in production?

(FWIW older versions of IE don't support `console.log()`; not sure if that's a concern for us or not by now)

nosegraze: I'd like to keep it due to the huge number of _"registration form not working / hanging"_ tickets we get in support. I have no idea if it's ever actually been useful, but we've had it there until now so I think we might as well keep it.

tnorthcutt: Makes sense! Adding back.

tnorthcutt: Got it!

tnorthcutt:
I think this is ready to merge in for `2.9.11`

PR #1743, opened 2018-03-29
Display consistent currency symbol in settings

### LINKED ISSUE #105: Currency chosen should dynamically change the currency position select menu
Choosing a currency that uses a symbol other than "$" should dynamically update the "Currency Position" select list to use the appropriate symbol, if applicable.

I'm assuming the simplest (best?) way to do this is to store an array of possible "before/after" text, and use js to swap out them out if the selected currency matches one of the before/after pairs. Hopefully that makes some sense.

By the way, I realize this is a tiny, tiny enhancement that impacts the user for all of probably five seconds, so I won't feel bad at all if this is lower than low on the priority list :smile:.

2014-08-07 at 11 06 pm

tnorthcutt:
When a different currency is chosen, this dynamically changes the currency symbol in the Currencty Position select element to match.

Closes #105

### DIFF
--- a/includes/js/admin-scripts.js
+++ b/includes/js/admin-scripts.js
@@ -356,31 +356,33 @@ jQuery(document).ready(function($) {
 		}
 	});
 
-    var currencies = JSON.parse(rcp_vars.currencies);
     var currencySelect = $('#rcp_settings\\[currency\\]');
-    var currentsymbol = currencies[currencySelect.val()].match(/\(([^)]+)\)/)[1];
-    var before = $('#rcp_settings\\[currency_position\\] option[value="before"]');
-    var after = $('#rcp_settings\\[currency_position\\] option[value="after"]');
-
-    before.text(function () {
-        return $(this).text().replace("$", currentsymbol);
-    });
-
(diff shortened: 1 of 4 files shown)


includes/js/admin-scripts.js
```diff
@@ -356,4 +356,60 @@ jQuery(document).ready(function($) {
 		}
 	});
 
+    var currencies = {
```
pippinsplugins: Instead of repeating these, could we insert them through wp_localize_script() like the other PHP > JS variables?

tnorthcutt: I think so, yes. I initially shied away from that because we only need the three letter code and the symbol (and not e.g. "US Dollars"), but after looking at it again, since the symbol entity is inside parentheses I think I can extract it.

tnorthcutt:
@pippinsplugins Ok, ready to go now 😁

tnorthcutt:
@mindctrl 👍

mindctrl:
Latest changes look good. I'm going to fix the indentation problems (convert spaces to tabs), add some docs, and merge.

PR #1747, opened 2018-03-29
Expand invoice detail

### LINKED ISSUE #1150: Expand invoice detail to show fees/discounts/credits
After #1000 I'd love to see more detail put into the invoices.

In one of our meetings we talked about the possibility of adding more payment columns for all the possible discounts/fees, and the idea is really growing on me. That could tie in nicely to expanding the invoices. If we added database columns for like `proration_credits`, `fees`, and maybe `discounted_amount`, we could include that kind of information on the invoices.

I got a ticket from a customer who had someone signup for a £40 plan with a discount code and the invoice just showed £20. He requested that the original £40 amount be shown, along with the amount that was discounted.

## Current
rcp-invoice-current

## Proposed
rcp-invoice-proposed

tnorthcutt:
Now shows a subscription level's price, followed by any fees, then the resulting subtotal, then any credits and discounts, then the total charge.

Closes #1150


tnorthcutt:
@mindctrl ok, I think this is much better. Please double check my logic if you don't mind 😄 (see [commit message](https://github.com/restrictcontentpro/restrict-content-pro/pull/1747/commits/479e63abb6e0dca411bc44b97f2921e53250aee4) for an explanation).

mindctrl:
@tnorthcutt backfill logic looks right to me now. I think all that's left is not showing 0 value rows per https://github.com/restrictcontentpro/restrict-content-pro/issues/1150#issuecomment-378764037

templates/invoice.php
```diff
@@ -265,23 +268,53 @@
 		<!-- Items -->
 		<section id="items">
 
-			<header><?php _e( 'Invoice Items:', 'rcp' ); ?></header>
+			<header><strong><span class="alignleft"><?php _e( 'Description', 'rcp' ); ?></span><span class="right"><?php _e( 'Amount', 'rcp' ); ?></span></strong></header>
```
nosegraze: Instead of keeping this in a `<header>` tag we should remove the `<header>` tag completely and put these headings inside the table in `<thead>`.

nosegraze:
Just a small markup change, then I'm happy with it. 👍

tnorthcutt:
FYI I implemented this as e.g. `echo rcp_currency_filter( -1 * abs( $rcp_payment->credits ) );` so that if we store credits and/or discounts as negatives in the future, the display code will still work.

Issue #1711, opened 2018-03-07
Custom table for subscriptions


mindctrl:
I'm breaking this out separate from #774.

This may be updated, but the current schema proposal is:

Field | Type | Null | Key | Default | Extra
-- | -- | -- | -- | -- | --
id | int unsigned | NO | PRI | NULL | auto_increment
object_id | bigint unsigned | NO | MUL | NULL |
object_type | varchar(20) | NO |   | 'subscription' | 
user_id | int unsigned | NO | MUL | NULL |
currency | varchar(20) | NO | | USD |
initial_amount | mediumint | NO | | NULL |
recurring_amount | mediumint unsigned | NO | | NULL |
created_date | datetime | NO | | CURRENT_TIMESTAMP |
trial_end_date | datetime | YES | | NULL |
cancellation_date | datetime | YES | | NULL |
expiration_date | datetime | YES | | NULL |
times_billed unsigned | smallint | NO | NO | 1 | |
status | varchar(12) | NO | MUL | active |
payment_profile_id | tinytext | YES | | NULL |
gateway | tinytext | NO | | NULL |

nosegraze:
Is it worth renaming `payment_profile_id` to something a little more descriptive? Like `gateway_subscription_id`? Or `gateway_profile_id`? Just something to more clearly designate that it's a gateway value.

Also, how does `expiration_date` (and probably `cancellation_date` too) work? For example, say a subscription expires, so we fill out `expiration_date`, but then it gets renewed and reactivated. Is `expiration_date` still there or do we wipe it out? Then do we overwrite the value if the subscription expires yet again?

Also what about a column for `auto_renew` or `recurring`?

Also, now that I'm thinking about auto renew, I think we once vaguely discussed removing the `cancelled` status completely and instead thinking of cancellations more in terms of _turning off auto renew_ rather than cancelling membership. Thoughts on that?

nosegraze:
Lol don't mind me. Yesterday I seem to have forgotten how expiration dates work.

nosegraze:
I think I thought `times_billed` was going to be the current number of times the subscription _has been_ billed - not the number of times a subscription _will be_ billed. So you sign up, that's `1`, then you renew and that makes it `2`, etc.

mindctrl:
[...]

PR #1750, opened 2018-04-03
Add ability to specify subscription level(s) for notices

tnorthcutt:
Adds checkboxes of subscription levels when adding or editing a renewal or expiration notice. If any levels are checked, then the notice will only be sent to members with those subscription levels.

If no boxes are checked, the notice will be sent to all applicable members regardless of subscription level.


includes/admin/reminders/subscription-reminder-view.php
```diff
@@ -80,6 +81,30 @@
...
+			<tr>
+				<th scope="row" valign="top">
+					<label for="rcp-notice-levels"><?php _e( 'Membership Levels', 'rcp' ); ?></label>
+				</th>
+				<td>
+					<?php
+					$levels = rcp_get_subscription_levels( 'all' );
+					if ( $levels ) : ?>
+						<?php
+						$current = maybe_unserialize( $notice['levels'] );
```
nosegraze: The notices are already stored as an array in the options table, so the `maybe_unserialize()` shouldn't be necessary.

includes/admin/reminders/subscription-reminder-view.php
```diff
@@ -80,6 +81,30 @@
...
+						if ( ! isset( $current ) || ( is_array( $current ) && empty( $current ) ) || ! $current ) {
```
nosegraze: This could be simplified to:

```php
if ( empty( $current ) || ! is_array( $current ) )
```

includes/admin/reminders/subscription-reminder-view.php
```diff
@@ -80,6 +81,30 @@
...
+							$current = array();
+						}
+						foreach ( $levels as $level ) :
+						?>
+							<input type="checkbox" id="rcp-notice-levels-<?php echo $level->id; ?>" name="rcp_reminder_levels[]" value="<?php echo $level->id ?>" <?php checked( true, in_array($level->id, $current ) ); ?>>
+							<label for="rcp-notice-levels-<?php echo $level->id; ?>"><?php echo $level->name; ?></label>
```
nosegraze: Please add:

`esc_attr()` on the input ID and value, and the label `for` attribute.
`esc_html()` on the level name.

includes/admin/reminders/subscription-reminders.php
```diff
@@ -60,6 +61,18 @@ function rcp_subscription_reminder_table( $type = 'expiration' ) {
 					<td><?php echo esc_html( stripslashes( $notice['subject'] ) ); ?></td>
 					<td><?php echo esc_html( $reminders->get_notice_period_label( $key ) ); ?></td>
 					<td><?php echo ! empty( $notice['enabled'] ) ? __( 'Enabled', 'rcp' ) : __( 'Disabled', 'rcp' ); ?></td>
+					<td>
+						<?php
+						$levels = maybe_unserialize( $notice['levels'] );
```
nosegraze: Shouldn't need `maybe_unserialize()` as `get_option()` already does that.

includes/admin/reminders/subscription-reminders.php
```diff
@@ -103,6 +116,7 @@ function rcp_process_add_edit_reminder_notice() {
 	$message   = isset( $_POST['rcp_reminder_message'] ) ? wp_kses( stripslashes( $_POST['rcp_reminder_message'] ), wp_kses_allowed_html( 'post' ) ) : false;
 	$type      = isset( $_POST['rcp_reminder_type'] ) ? sanitize_text_field( $_POST['rcp_reminder_type'] ) : 'renewal';
 	$enabled   = isset( $_POST['rcp_reminder_enabled'] );
+	$levels    = isset( $_POST['rcp_reminder_levels'] ) ? maybe_serialize( $_POST['rcp_reminder_levels'] ) : '';
```
nosegraze: Shouldn't need `maybe_serialize()` as `update_option()` already handles that later.

includes/class-rcp-reminders.php
```diff
@@ -257,7 +257,20 @@ public function send_reminders() {
 					continue;
 				}
 
-				$members = $this->get_reminder_subscriptions( $notice['send_period'], $type );
+				$levels = maybe_unserialize( $notice['levels'] );
```
nosegraze: Don't need `maybe_unserialize()`.

includes/admin/reminders/subscription-reminders.php
nosegraze: Also we need to sanitize this value before we save it. `array_map()` `absint()`.

includes/admin/reminders/subscription-reminders.php
nosegraze: Also `$notice['levels']` will not be set yet when upgrading so we need a check to see if it actually exists before trying to use it.

includes/admin/reminders/subscription-reminder-view.php
nosegraze: Also `$notice['levels']` will not be set yet when upgrading so we need a check to see if it actually exists before trying to use it.

includes/class-rcp-reminders.php
nosegraze: We also need to check if the `levels` key exists before using it. It won't exist on old reminders.

includes/admin/reminders/subscription-reminders.php
tnorthcutt: 👍

includes/admin/reminders/subscription-reminder-view.php
tnorthcutt: 👍

includes/admin/reminders/subscription-reminder-view.php
tnorthcutt: 👍

includes/admin/reminders/subscription-reminder-view.php
tnorthcutt: 👍

includes/admin/reminders/subscription-reminder-view.php
tnorthcutt: Turns out if we skip `maybe_unserialize()` here, we just get a string like `a:1:{i:0;s:1:"5";}` instead of an actual array... so I'm going to add back in our serializing and unserializing.

includes/admin/reminders/subscription-reminders.php
tnorthcutt: 👍

includes/class-rcp-reminders.php
tnorthcutt: 👍

tnorthcutt:
@nosegraze I had to do that separately from the main logging line there, because at that point we don't yet have the levels. Let me know what you think.

includes/admin/reminders/subscription-reminder-view.php
nosegraze: Hm it works fine for me without unserializing. It was coming out like that for you when using `get_option()` to retrieve it? It comes out unserialized for me.

This is a `var_dump()` on `get_option( 'rcp_reminder_notices' )` : 

```php
    array (size=6)
      'subject' => string 'Your Subscription is About to Expire' (length=53)
      'message' => string 'Hello %name%,

Your subscription for %subscription_name% will expire on %expiration%.' (length=87)
      'send_period' => string '+1day' (length=5)
      'type' => string 'expiration' (length=10)
      'enabled' => boolean false
      'levels' => 
        array (size=2)
          0 => int 1
          1 => int 2
```

tnorthcutt: @nosegraze I swear I got a string at one point... but I can't reproduce it now ¯\\\_(ツ)_/¯. We'll go with "you were right the whole time and I'm crazy" 😁

tnorthcutt:
@nosegraze yeah fair enough on fetching `$levels` earlier. I'd originally not done that since we have a couple of early return checks for `enabled` and empty message/subject, but I suppose that wasn't saving us much to wait on getting the levels.

Moved it ⬆️

nosegraze:
```php
rcp_log( sprintf( 'Processing %s reminder. ID: %d; Period: %s. Levels: %s', $type, $notice_id, $notice['send_period'], $levels ) );
```

Can you please put `$levels` in `var_export()` like this:

```php
rcp_log( sprintf( 'Processing %s reminder. ID: %d; Period: %s. Levels: %s', $type, $notice_id, $notice['send_period'], var_export( $levels, true ) ) );
```

Just because it's an array and that'll basically convert it to a string.

tnorthcutt:
@nosegraze ah geez, you even said that earlier. My bad. Fixed now!

nosegraze:
Almost - you need to add a second parameter in `var_export()` that's `true` so that it's returned instead of echoed. Like this:

```php
var_export( $levels, true )
```

Thanks!

nosegraze:
I'm happy with this now, but going to let the PR sit until after https://github.com/restrictcontentpro/restrict-content-pro/issues/1711 because we'll need to edit the query once the table gets added in.

tnorthcutt:
@nosegraze if #1711 isn't going to be released soon, is there a downside (for RCP customers) to merging this now? Meaning, would the UI change, or would they lose any saved notices once #1711 gets done?

If not, perhaps we could reconsider adding this in now. I know of at least one client who wants to send renewal notices to annual members but not monthly members, and of course can't right now. Perhaps there's a downside I'm not thinking of?

Issue #1755, opened 2018-04-04
Add a new member signup method option for "Imported"


nosegraze:
On the "Edit Member" screen we support two signup methods in the dropdown:

1. User Signup
2. Added by admin manually

It would be useful to add a third option for `Imported` to designate they were imported via CSV. This would be helpful in support too if you need to do some debugging that involves figuring out how the user was actually given their subscription.

tnorthcutt:
Quick note: this will require a corresponding change to the [RCP CSV User Import](https://github.com/restrictcontentpro/rcp-csv-user-import) addon; I'll make that PR (hopefully immediately) after this one.

nosegraze:
PR: #1761

PR #1761, opened 2018-04-07
Add "imported" option for new user signup method

tnorthcutt:
Closes https://github.com/restrictcontentpro/restrict-content-pro/issues/1755

N.B. I'm not stuck on the wording of the option; for now it's:
> Imported (usually by CSV)

...but I'm open to changing it. I figured that's somewhat safe, since if someone programmatically creates new users they might want to use the new `imported` method, even if it's not done with a CSV.


nosegraze:
I'm inclined to just stick with a vague `Imported`, but I'm not really fussed either way.

mindctrl:
I like `Imported` too. I don't think we need the added wording of "usually by CSV".

tnorthcutt:
Fine by me 😄. Updated.

PR #1762, opened 2018-04-09
Prevent saving of post if conflicting RCP settings

### LINKED ISSUE #1706: Prevent saving post and/or show error message if selecting "members of specific levels" with no levels selected
- Select `Members of subscription level(s)`.
- Select `Members of specific subscription levels`.
- Do not check any levels.

A customer noted that forgetting to check a level before saving would result in the post being completely public. We should consider showing an error message and maybe even prevent the post from being saved until the settings are completed (or the option changed).

tnorthcutt:
If "Members of specific subscription levels" is selected but no subscription levels are checked, this will show an alert box and prevent saving of the post.

The script localization of the messages is a little messy, but that's the only way I could come up with to have a multiline message that still gets translated.

I feel the "Restrict Content Pro message:" in the alert box is important so the user knows where this is coming from/has some context.

Closes #1706


includes/js/admin-scripts.js
```diff
@@ -356,4 +356,18 @@ jQuery(document).ready(function($) {
 		}
 	});
 
+	$('#publish').click(function () {
+		if ($('#rcp_subscription_level_specific').is(':checked')) {
+			let levels = $('.rcp-subscription-levels').find('input.rcp_subscription_level');
+
+			if ($(levels).is(':checked')) {
```
nosegraze: Let's simplify this by just doing:

```js
if ( ! $(levels).is(':checked') )
```

includes/scripts.php
```diff
@@ -42,19 +42,21 @@ function rcp_admin_scripts( $hook ) {
...
+				'delete_reminder'         => __( 'Are you sure you want to delete this reminder email? This action is irreversible. Proceed?', 'rcp' ),
+				'missing_username'        => __( 'You must choose a username', 'rcp' ),
+				'currency_sign'           => rcp_currency_filter( '' ),
+				'currency_pos'            => isset( $rcp_options['currency_position'] ) ? $rcp_options['currency_position'] : 'before',
+				'use_as_logo'             => __( 'Use as Logo', 'rcp' ),
+				'choose_logo'             => __( 'Choose a Logo', 'rcp' ),
+				'can_cancel_member'       => ( $hook == $rcp_members_page && isset( $_GET['edit_member'] ) && rcp_can_member_cancel( absint( $_GET['edit_member'] ) ) ),
+				'cancel_subscription'     => __( 'Cancel subscription at gateway', 'rcp' ),
+				'rcp_alert_message'       => __( 'Restrict Content Pro message:', 'rcp' ),
+				'specific_levels_message' => __( 'Because "Members of specific subscription levels" is chosen, you must check at least one subscription level in order to save this content.', 'rcp' )
```
nosegraze: I don't love starting this sentence with _"Because"_. I'm open to other thoughts but I might word it more like this:

> You have restricted this content to "Members of specific subscription levels" but have not selected any levels. Please check at least one subscription level in order to save this content.

includes/js/admin-scripts.js
```diff
@@ -356,4 +356,18 @@ jQuery(document).ready(function($) {
 		}
 	});
 
+	$('#publish').click(function () {
```
nosegraze: This doesn't factor in saving as a draft. Will write a more detailed note about this.

mindctrl:
Does this work in Gutenberg? I just tried to load Gutenberg but it's not working on my dev site.

mindctrl:
Confirmed it does not work with Gutenberg. The alert does not pop.

nosegraze:
Yes I like the idea of a query var / admin notice. We can simply not save the restriction at all if there are no levels selected and show an admin notice saying that it wasn't saved and explain why.

includes/scripts.php
tnorthcutt: 👍

includes/js/admin-scripts.js
tnorthcutt: This is moot since we're no longer using a js alert for the message.

includes/js/admin-scripts.js
tnorthcutt: This is moot since we're no longer using a js alert for the message.

tnorthcutt:
Ok, I pushed some changes so this no longer uses a js alert to warn.

It still doesn't work on Gutenberg (surprise surprise, admin notices are totally different for Gutenberg 🙄), but I'm looking into it.

tnorthcutt:
Apparently there is not an actual API for notices in Gutenberg yet: https://github.com/WordPress/gutenberg/issues/5975

includes/admin/metabox.php
```diff
@@ -151,6 +151,13 @@ function rcp_save_meta_data( $post_id ) {
 
 					$levels = array_map( 'absint', $_POST[ 'rcp_subscription_level' ] );
 
+					if( $levels === null ) {
+						add_filter( 'get_edit_post_link', function ( $link ) {
+							$link = add_query_arg( [ 'rcp_mismatch' => 'true' ], $link );
```
nosegraze: The short array syntax is PHP `5.4+`. We unfortunately support `5.3` so this needs to be changed.

nosegraze:
This still results in a bunch of PHP warnings when saved:

```
PHP Notice:  Undefined index: rcp_subscription_level in C:\wamp\www\rcp\wp-content\plugins\restrict-content-pro\includes\admin\metabox.php on line 152
PHP Warning:  array_map(): Argument #2 should be an array in C:\wamp\www\rcp\wp-content\plugins\restrict-content-pro\includes\admin\metabox.php on line 152
PHP Warning:  Invalid argument supplied for foreach() in C:\wamp\www\rcp\wp-content\plugins\restrict-content-pro\includes\admin\metabox.php on line 161
```

We'll want to change that block to something like this:

```php
case 'specific' :

	$levels = ! empty( $_POST[ 'rcp_subscription_level' ] ) ? $_POST[ 'rcp_subscription_level' ] : array();
	$levels = array_map( 'absint', $levels );

	if( empty( $levels ) ) {
		add_filter( 'get_edit_post_link', function ( $link ) {
			$link = add_query_arg( [ 'rcp_mismatch' => 'true' ], $link );
			return $link;
		} );

		break;
	}
```

Additionally, now that we are letting the post fundamentally be saved, the message will probably need to be tweaked again. Right now it's:

> You have restricted this content to "Members of specific subscription levels" but have not selected any levels. Please check at least one subscription level in order to save this content.

We might want to change it to this:

You have restricted this content to "Members of specific subscription levels" but have not selected any levels. Please check at least one subscription level in order to restrict this content.

(just one small word edit)

Because now it has actually been saved - just maybe not how they expect.

tnorthcutt:
@nosegraze thanks for the improvements.

This still doesn't work with Gutenberg. I haven't been able to sort that out yet, but I'll come back to it in the near-ish future to poke at it some more.

nosegraze:
I'm fine with holding off on Gutenberg for now and revisiting it later down the line.

Can the admin notice be tweaked to follow the same format as the rest of the notices? Change the query arg to something like `rcp_message=invalid-restriction`, populate the `$text` variable with the message, and change `$class = 'error';`

That will keep it consistent with all the other notices.

tnorthcutt:
@nosegraze ah yes, that's much cleaner! Fixed now.

mindctrl:
@tnorthcutt Keeping the previous restrictions, if any, would prevent any accidental content exposure. I like it. 👍

tnorthcutt:
@mindctrl ok, think this is good now. Now if "specific levels" is chosen but no levels are checked, all restriction saving is aborted before any changes are recorded.

PR #1775, opened 2018-04-12
Rework PayPal Express confirmation screen #1744

### LINKED ISSUE #1744: One-time discounts with auto renew not reflected on PayPal Express confirmation page
Haven't tested/confirmed this yet, but got a report that one-time discounts weren't reflected on the PayPal Express confirmation page. It was only showing the recurring amount - not the other, lower amount actually being paid today. A quick look at the code indicates that's probably correct. This customer was using auto renew.

nosegraze:
Closes #1744


nosegraze:
Updated 👍

mindctrl:
@nosegraze looks better! We might want to rename those new classes. "name" and "price" are pretty generic, and even though we're scoping properly in our CSS, I imagine some other software doesn't.

Something like `rcp-ppe-confirm-price`?

nosegraze:
Yes, good point. I copied the class names from `invoice.php` without really thinking, but prefixing is much more important when `wp_head` is actually being included.

PR #1777, opened 2018-04-12
[WIP] use WP_List_Table for members admin page

### LINKED ISSUE #1685: Use WP_List_Table on the Members page listing
Right now we have a custom page that basically replicates all the functionality in `WP_List_Table`. Maybe we should consider converting this page to extend it. We'd gain any improvements they make to it.

Related: #101, #1262

tnorthcutt:
**Still a work in progress; not ready for review.**

Some of this is a bit hacky; for instance I just created a new submenu page instead of replacing the current one for now, so I can use the current one as a reference.

This borrows heavily from https://github.com/Veraxus/wp-list-table-example.

Closes #1685 when complete.


tnorthcutt:
Just an update (not ready for merge yet): this is basically working as-is now. I still want to clean a lot of it up and think through what can be better abstracted out to our base `RCP_Admin_List_Table` class, but I'm happy with it so far.

Once complete, this will spawn new issues for moving Discounts and Payments to a WP list table as well.

tnorthcutt:
Something to note:

Currently, we run an `rcp_member_row_actions` action immediately after the row actions div in the members table. However, part of this PR moves us to using `WP_List_Table::row_actions()`, which does not have a corresponding action we could piggyback on.

Consequently, this will deprecate `rcp_member_row_actions`. I don't know how widely used that action is, but it's definitely something to note.

For instance, it is used in the WP Approve User integration which ships with RCP.

One option to eliminate this problem would be to override `WP_List_Table::row_actions()` in either `RCP_Admin_List_Table` or `RCP_Members_List_Table` so we can retain that action.

I'd like input from others on the preferred way to handle this. /cc @pippinsplugins @nosegraze @mindctrl

tnorthcutt:
@nosegraze one issue with that: multiple things hooked to `rcp_member_row_actions` would not result in multiple elements added to the `$actions` array. Instead, they'd all be added to one `<span>`, and not have separators `|` added.

I wonder if there is some way to iterate over the things hooked there...?

nosegraze:
It's always been up to individuals to add their own separators, and if they added in the HTML for them then they would appear. As I said, it may not look perfect design-wise but I think it should still function perfectly fine, which is really the only important thing. As long as it functions I don't think we need to invest a ton of time in making it look perfect.

tnorthcutt:
@nosegraze oh, of course it is – my  bad! Thanks for pointing that out.

nosegraze:
I'm not bothered by a double separator and I would be fine with just leaving it as-is. This action would be deprecated and we'd probably have some new filter (unless there's already one in core?) for hooking in for the future, and people would be encouraged to update their code accordingly. So only legacy code would have a double separator and that's really just a minor display quirk.

tnorthcutt:
Got it 👍

(There isn't a filter in core, FWIW)

tnorthcutt:
Dropping the WIP designation as this is in much better shape now; I'd be happy for a preliminary review if anyone wants to take a look.

includes/admin/admin-actions.php
```diff
@@ -23,6 +23,10 @@ function rcp_process_actions() {
 		do_action( 'rcp_action_' . $_POST['rcp-action'], $_POST );
 	}
 
+	if ( isset( $_POST['action'] ) ) {
```
pippinsplugins: Can we avoid this? This will cause us to fire a do_action anytime anyone sens a post with a key of `action`, which is super common. That's too over reaching in my opinion.

includes/admin/class-rcp-admin-list-table.php
```diff
@@ -0,0 +1,196 @@
+<?php
+
+if (! class_exists('WP_List_Table')) {
```
pippinsplugins: Need a space before `!`

includes/admin/class-rcp-admin-list-table.php
```diff
@@ -0,0 +1,196 @@
...
+	require_once ABSPATH . 'wp-admin/includes/class-wp-list-table.php';
+}
+
+/**
+ * RCP_Admin_List_Table
+ *
+ * Base list table class for admin list tables.
+ *
+ * @package     Restrict Content Pro
+ * @copyright   Copyright (c) 2017, Pippin Williamson
```
pippinsplugins: Date here, and elsewhere, should be `2018`.

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+			$this->active_count    = rcp_get_member_count( 'active' );
+			$this->pending_count   = rcp_get_member_count( 'pending' );
+			$this->expired_count   = rcp_get_member_count( 'expired' );
+			$this->cancelled_count = rcp_get_member_count( 'cancelled' );
+			$this->free_count      = rcp_get_member_count( 'free' );
+			$this->total_count     = $this->active_count + $this->pending_count + $this->expired_count + $this->cancelled_count + $this->free_count;
+		}
+	}
+
+	public function get_columns() {
```
pippinsplugins: Needs PHPDoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+			'status'       => _x( 'Status', 'Column label', 'rcp' ),
+			'recurring'    => _x( 'Recurring', 'Column label', 'rcp' ),
+			'expiration'   => _x( 'Expiration', 'Column label', 'rcp' ),
+			'user-role'    => _x( 'User Role', 'Column label', 'rcp' ),
+		);
+
+		return $columns;
+	}
+
+	protected function get_sortable_columns() {
```
pippinsplugins: Needs PHPdoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		);
+
+		return $sortable_columns;
+	}
+
+	/**
+	 * Retrieve the view types
+	 *
+	 * @access public
+	 * @since 1.4
```
pippinsplugins: Version since should be 3.0

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+			'pending'   => sprintf( '<a href="%s"%s>%s</a>', add_query_arg( array( 'status' => 'pending', 'paged'  => false ) ), $current === 'pending' ? ' class="current"' : '', __( 'Pending', 'rcp' ) . $pending_count ),
+			'expired'   => sprintf( '<a href="%s"%s>%s</a>', add_query_arg( array( 'status' => 'expired', 'paged'  => false ) ), $current === 'expired' ? ' class="current"' : '', __( 'Expired', 'rcp' ) . $expired_count ),
+			'cancelled' => sprintf( '<a href="%s"%s>%s</a>', add_query_arg( array( 'status' => 'cancelled', 'paged'  => false ) ), $current === 'cancelled' ? ' class="current"' : '', __( 'Cancelled', 'rcp' ) . $cancelled_count ),
+			'free'      => sprintf( '<a href="%s"%s>%s</a>', add_query_arg( array( 'status' => 'free', 'paged'  => false ) ), $current === 'free' ? ' class="current"' : '', __( 'Free', 'rcp' ) . $free_count ),
+		);
+
+		return apply_filters( 'rcp_members_table_views', $views );
+	}
+
+	protected function column_default( $item, $column_name ) {
```
pippinsplugins: Needs PHPdoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		switch ( $column_name ) {
+			default:
+				return $item[$column_name];
+		}
+	}
+
+	protected function column_cb( $item ) {
```
pippinsplugins: Needs PHPdoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		return sprintf(
+			'<input type="checkbox" name="%1$s[]" value="%2$s" />',
+			$this->_args['singular'] . '-ids',  // Let's simply re-purpose the table's singular label ("member").
+			$item['ID']                // The value of the checkbox should be the record's ID.
+		);
+	}
+
+	protected function column_user( $item ) {
```
pippinsplugins: Needs PHPdoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+			esc_url( wp_nonce_url( add_query_arg( 'edit_member', $item['ID'], $current_page ) ) ),
+			esc_attr( 'Edit Member', 'rcp' ),
+			esc_html( $item['user'] ),
+			$item['user'] != $item['email'] ? esc_html( '&nbsp;&ndash;&nbsp;' . $item['email'] ) : '',
+			esc_html( $item['ID'] ),
+			$this->row_actions( $actions )
+		);
+	}
+
+	protected function get_bulk_actions() {
```
pippinsplugins: Needs PHPdoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		$actions = array(
+			'mark-active'    => _x( 'Mark as Active', 'List table bulk action', 'rcp' ),
+			'mark-expired'   => _x( 'Mark as Expired', 'List table bulk action', 'rcp' ),
+			'mark-cancelled' => _x( 'Mark as Cancelled', 'List table bulk action', 'rcp' ),
+		);
+
+		return $actions;
+	}
+
+	protected function bulk_actions( $which = '' ) {
```
pippinsplugins: Needs PHPdoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		}
+
+		echo "</select>\n";
+
+		?><input type="text" class="rcp-datepicker" name="expiration<?php echo $two; ?>" placeholder="<?php esc_attr_e( 'New Expiration Date', 'rcp' ); ?>" id="rcp-bulk-expiration" value=""/><?php
+		submit_button( __( 'Apply' ), 'action', '', false, array( 'id' => "doaction$two" ) );
+		echo "\n";
+	}
+
+	function prepare_items() {
```
pippinsplugins: Needs PhPdoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		) );
+	}
+
+	/**
+	 * Allows you to sort the data by the variables set in the $_GET
+	 *
+	 * @return Mixed
+	 */
+	private function sort_data( $a, $b )
+	{
```
pippinsplugins: This should be on the same line as the function definition.

pippinsplugins: This whole function, actually, needs to be reformatted to match our standards

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		}
+		$result = strcmp( $a[$orderby], $b[$orderby] );
+		if($order === 'asc')
+		{
+			return $result;
+		}
+		return -$result;
+	}
+
+	private function expiration_sort( $a, $b, $order ) {
```
pippinsplugins: Needs PHPDoc

includes/admin/members/class-rcp-members-list-table.php
```diff
@@ -0,0 +1,454 @@
...
+		$result = strcmp( $a_date, $b_date );
+
+		if ( $order === 'asc' ) {
+			return $result;
+		}
+
+		return -$result;
+	}
+
+	protected function get_table_classes() {
```
pippinsplugins: Needs PHPDoc

includes/admin/members/member-actions.php
```diff
@@ -230,7 +230,7 @@ function rcp_process_bulk_edit_members() {
 	}
 
 	$member_ids = array_map( 'absint', $_POST['member-ids'] );
-	$action     = ! empty( $_POST['rcp-bulk-action'] ) ? sanitize_text_field( $_POST['rcp-bulk-action'] ) : false;
+	$action     = ! empty( $_POST['action'] ) ? sanitize_text_field( $_POST['action'] ) : false;
```
pippinsplugins: Per my comment above regarding how `action` is over-reaching, can we prefix this like it used to be?

includes/admin/members/member-actions.php
```diff
@@ -239,6 +239,9 @@ function rcp_process_bulk_edit_members() {
 		if ( ! empty( $_POST['expiration'] ) && 'delete' !== $action ) {
 			$member->set_expiration_date( date( 'Y-m-d H:i:s', strtotime( $_POST['expiration'], current_time( 'timestamp' ) ) ) );
 		}
+		else if ( ! empty( $_POST['expiration2'] ) && 'delete' !== $action ) {
```
pippinsplugins: `}` and `else` should be on the same line.

includes/admin/class-rcp-admin-list-table.php
tnorthcutt: 👍

includes/admin/class-rcp-admin-list-table.php
tnorthcutt: 👍

includes/admin/members/class-rcp-members-list-table.php
tnorthcutt: 👍

includes/admin/members/class-rcp-members-list-table.php
tnorthcutt: 👍

includes/admin/members/member-actions.php
tnorthcutt: 👍

includes/admin/admin-actions.php
tnorthcutt: Yeah, makes sense. I think I can do away with that.

includes/admin/members/member-actions.php
tnorthcutt: 👍

includes/admin/admin-actions.php
tnorthcutt: Actually now that I look at it more, while we can do away with this, the "WordPress Way" is to use `action` (and `action2`): https://github.com/WordPress/WordPress/blob/cf0d00c602452fab849670c7deefadc3dca03875/wp-admin/includes/class-wp-list-table.php#L485

I'm not saying we need to do it that way, of course – but perhaps it should be considered?

Issue #1736, opened 2018-03-24
Move from post meta to custom taxonomies for restriction data


tnorthcutt:
_This is an offshoot of https://github.com/restrictcontentpro/restrict-content-pro/issues/135_

@mindctrl and I discussed this a bit today. Right now, `rcp_get_restricted_post_ids()` does a [pretty big post meta query](https://github.com/restrictcontentpro/restrict-content-pro/blob/a00264b2001c9553f3138b1dc478b4913ff327d2/includes/misc-functions.php#L1099-L1140), which on large sites can be a problem.

Some very rudimentary testing (calling `rcp_get_restricted_post_ids()` in my theme's `page.php` template) with Query Monitor confirmed this query gets very slow pretty quickly, relative to number of posts in the database. At a few thousand posts, this got up to .25 seconds.

I'll run some additional tests to try and get a rough comparison to a taxonomy query that would be roughly what we'd use, and update this issue.

tnorthcutt:
Update: some rudimentary testing suggests that with ~6k posts, using taxonomies instead of post meta results in about 1/5 the query time. Very rough, but a big enough difference to make further work on this worth exploring.

tnorthcutt:
On hold for now per https://github.com/restrictcontentpro/restrict-content-pro/issues/1670#issuecomment-375914648

tnorthcutt:
Revisiting this today, it seems my earlier conclusions about query performance were premature. I think I may have caused the initial testing to be slow by performing those queries immediately after using wp-cli to insert several thousand posts with associated post meta. My theory is that temporarily degraded my CPU performance which impacted the subsequent queries.

So far I haven't been able to reproduce the slow (~.25 second) queries using the (existing) postmeta method; today they're taking roughly the same amount of time as the taxonomy query method (~.05 seconds).

I'm still open to making this change in the interest of a cleaner data model, and perhaps those query times would be more impacted on a larger site with a five or six digit number of posts.

@mindctrl I'd love your input on this 🙂

mindctrl:
@tnorthcutt we really only see it on crappy hosts that limit cpu or throttle mysql in some way (I've seen Hostgator throttle MySQL so that it rejects connections and causes data loss on other non-RCP sites), and/or sites with the five/six digit post counts. You should see a performance hit after 10k posts or some random high number like that.

mindctrl:
[...]

PR #1786, opened 2018-04-18
WIP of moving restricted post query to taxonomies

tnorthcutt:
Not at all functional; these changes are only for testing. This commit _must_ be reviewed to ensure no unwanted changes sneak in long-term.

Closes https://github.com/restrictcontentpro/restrict-content-pro/issues/1736 when complete


tnorthcutt:
The following functions are deprecated (and AFAICT not applied anywhere):
- `rcp_display_message_to_editors()`
- `rcp_display_message_authors()`
- `rcp_display_message_to_contributors()`
- `rcp_display_message_to_subscribers()`
- `rcp_display_message_to_non_logged_in_users()`

I'm wondering if I should bother updating those to use the new taxonomies instead of postmeta, e.g. replacing 
```
$user_level = get_post_meta( $post->ID, 'rcp_user_level', true );
```
with
```
$user_level = wp_get_post_terms( $post->ID, 'rcp_user_level', array( 'fields' => 'names' ) );
```

includes/misc-functions.php
```diff
@@ -1192,3 +1214,48 @@ function rcp_log( $message = '', $force = false ) {
...
+ *
+ * Returns either a minimum access level needed to view a post, or an empty string.
+ *
+ * @param $post_id
+ *
+ * @return int|string
+ */
+function rcp_get_access_level_restrictions( $post_id ) {
+	$access_level = wp_get_post_terms( $post_id, 'rcp_access_level', array( 'fields' => 'names' ) );
+	$access_level = is_array( $access_level ) ? $access_level[0] : '';
```
mindctrl: This throws an undefined offset notice with the following config.

image

includes/misc-functions.php
```diff
@@ -1106,32 +1106,54 @@ function rcp_get_restricted_post_ids() {
...
-		$post_ids = get_posts( array(
-			'post_status'    => 'publish',
-			'posts_per_page' => -1,
-			'post_type'      => 'any',
-			'fields'         => 'ids',
-			'meta_query'     => array(
-				'relation' => 'OR',
+	$post_ids = get_posts( array(
+		'post_status'    => 'publish',
+		'posts_per_page' => -1,
```
mindctrl: We'll need to do this via batching after #1612 is merged. Will ping here again when that's done.

includes/misc-functions.php
```diff
@@ -1192,3 +1214,48 @@ function rcp_log( $message = '', $force = false ) {
...
+ * Gets user role restrictions for a post
+ *
+ * Returns either a user role e.g. 'subscriber' or an empty string ''
+ *
+ * @param $post_id
+ *
+ * @return string
+ */
+function rcp_get_user_role_restrictions( $post_id ) {
+	$user_role = wp_get_post_terms( get_the_ID(), 'rcp_user_level', array( 'fields' => 'names' ) );
```
mindctrl: `get_the_ID()` should only be used here if `$post_id` is empty.

mindctrl: Actually not sure it should ever be used. We should be explicit about what post ID gets used to avoid unintended bugs.

mindctrl: After making that change, we also need to check that `$user_role` isn't empty, or the next line will throw an undefined offset notice because it'll be `null`.

mindctrl:
Edited the list above to add:

Filter `get_post_meta()` calls for backwards compatibility

mindctrl:
Since there are so many new methods being added to `RCP_Upgrades` here, I'd like to consider moving them out to another file or class so we don't muddy up the main upgrader class with a bunch of methods.

mindctrl:
@tnorthcutt have you pushed all your code up on this? I'm going to go ahead and take this and run with it, implementing batching, etc. and want to make sure I had all your work beforehand.

tnorthcutt:
@mindctrl Yep, that branch is updated. Have at it 👍

mindctrl:
Got it. Thanks @tnorthcutt !

PR #1815, opened 2018-05-15
Add terms & privacy checkboxes to registration form. #1814

### LINKED ISSUE #1814: Add optional terms and conditions / privacy policy checkboxes to registration form
Right now a terms and conditions checkbox is available via an add-on. Due to GDPR maybe we should put these in core.

nosegraze:
Closes #1814


nosegraze:
@mindctrl I don't mind. The only downside is that I've read `target="_blank"` isn't great for accessibility. But I guess it could be up to the end user to add _"(opens in new window)"_ to the label (which I've seen recommended).

mindctrl:
@nosegraze ah yes, that's mentioned here: https://www.w3.org/TR/WCAG20-TECHS/G201.html

If you agree, I say let's add it and if people mention accessibility let's link them to it. I think if we don't, it'll be a worse user experience and we'll get more tickets about people losing data entered in the form than we will about accessibility.

PR #1825, opened 2018-05-25
Only block access to content via REST API if user can't access it

### LINKED ISSUE #1281: Rest API Commenting issue
I have a site which uses the REST API to allow people to make comments. I've made the required changes in the admin, to allow "member" users to post comments. However, `rcp_post_password_required_rest_api` seems to put a blanket password restriction on all posts. This results in comments posted by members without the `edit_post` capability to fail.

Unregistering this hook does fix the problem, but that results in all the protected posts showing up in the REST API.

markjaquith:
fixes #1281


mindctrl:
Thanks @markjaquith!

PR #1840, opened 2018-06-20
Add a constant to disable email sending

### LINKED ISSUE #1839: Suppress all emails
I'd love a simple programmatic way to make sure RCP never sends an email.

We have a client where we're using the "restrict content" bit of RCP but not the account management/billing/etc (we are SSO'ing and syncing the member status data from a third party service… an increasingly common application, we're finding). This has me mildly paranoid that RCP will send a billing flow email when we don't want it to.

I'd love if there were a  way to just disable these across the board, perhaps something like `if ( defined( 'RCP_DISABLE_EMAILS' ) && RCP_DISABLED_EMAILS ) { return; }` in `RCP_Emails->send()`.

The reasons to do it this way, over disabling in the admin, are

1. Disabling each email in the admin is tedious
2. If new email notifications are added in the future, they don't default to disabled, so you could inadvertently start sending emails without realising it

Another option would be to allow filtering the `to` value on an email and just replace it with an admin email or some fake email. It's possible to do that with the `wp_mail` filter but that assumes that the filter is available… a bit risky because the `wp_mail` function is pluggable.

A third option would be to make this an admin setting but I feel like that's a dangerous option for RCP's non-technical users.

Happy to submit a PR, if this is of interest.

chrisvanpatten:
This PR introduces the `RCP_DISABLE_EMAILS` constant. When set to true, this will return early in the `RCP_Emails->send()` method to prevent emails from being sent.

This is currently set to return `true`, so RCP thinks the email was sent and you don't see "The e-mail could not be sent" errors.

Fixes #1839.


includes/class-rcp-emails.php
```diff
@@ -292,6 +292,10 @@ public function build_email( $message ) {
 	 */
 	public function send( $to, $subject, $message, $attachments = '' ) {
 
+		if ( defined( 'RCP_DISABLE_EMAILS' ) && RCP_DISABLE_EMAILS ) {
+			return true;
```
nosegraze: Let's add an `rcp_log()` right above here like so:

```php
rcp_log( 'Email not sent - detected RCP_DISABLE_EMAILS constant.', true );
```

chrisvanpatten:
I added the admin notice to `admin/admin-notices.php` because that seemed like the right place to do it. Confined to the settings page by checking `get_current_screen()`.

Here's how it looks:

restrict_content_pro_settings_ _mindful_ _wordpress_and_1__m___1___vim__tmux_

`notice-info` felt like the right level for this but it could also be `notice-error`… up to you! Also happy to tweak the message.

PR #1863, opened 2018-07-25
Add timestamp to the system info

### LINKED ISSUE #1862: Add "Date Generated" to system info file
Saw this in the EDD repo and would love it in RCP too.

> I would like to see a Date Generated: yyyy-mm-dd H:i:s added to the bottom of the system info file.

> Occasionally we get old copies of the system info file that makes the file not very helpful. By adding the date the file was generated, support techs can know for certain if it's an up to date version of the system info.

chrisvanpatten:
Fixes #1862 by adding a timestamp at the bottom of the system info file with the date/time the file is generated.

tools_ _mindful_ _wordpress

(Note that this info is also available in the date/time format setting under the "WordPress Configuration" section as well, but I think it's nice to have at the bottom too!)

### DIFF
--- a/includes/admin/tools/system-info.php
+++ b/includes/admin/tools/system-info.php
@@ -349,7 +349,10 @@ function rcp_tools_system_info_report() {
 
 	$return = apply_filters( 'rcp_system_info_after_php_extensions', $return );
 
-	$return .= "\n" . '### End System Info ###';
+	$return .= "\n" . '-- Miscellaneous' . "\n\n";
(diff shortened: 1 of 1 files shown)


nosegraze:
I'm actually tempted to say let's make the time in GMT. It won't make a huge difference in terms of date but then we always know what timezone it's in.

chrisvanpatten:
@nosegraze Great point — I thought I had set it to GMT but realised my test site was just set to GMT already!

Also simplified a bit: `current_time( 'mysql' )` will return in the date format we want anyway, so no need to pass it through `date()`.

nosegraze:
Thank you @chrisvanpatten ! :)

